PR #86, opened 2022-07-06
fix: stop select dropdown from throwing error when starting to type

dated:
## Summary

https://app.clickup.com/t/2tk8kw9

## Checklist

-   [x] Ready to be merged


goga-m (commented):
@dated all the fields work nicely now, We might need to change https://github.com/ArdentHQ/arkvault/blob/fix/select-error/src/domains/setting/pages/General/General.tsx#L326 
into `										if (marketProvider?.unsupportedCurrencies?.includes(exchangeCurrency)) {
`  because it will still throw if marketProvider is undefined/null

Other than it works great @dated :+1:

goga-m (approved):

samharperpittam (approved):

PR #129, opened 2022-08-09
refactor: implement message signing and verification as standalone pages

dated:
## Summary

https://app.clickup.com/t/2t3mf4b

## Checklist

-   [x] Ready to be merged


samharperpittam (changes requested):
Image should be hidden at breakpoint 375

Implementation
image

Design
image

Think the header text is the incorrect font size too. Is 12px in design but 16px in implementation

samharperpittam (changes requested):
Different error image

Implementation
image


Design
image

dated:
@samharperpittam, i fixed the image not being hidden. generic adjustments like the font size and image will be looked at in a separate PR, as they effect the whole application.

samharperpittam:
> @samharperpittam, i fixed the image not being hidden. generic adjustments like the font size and image will be looked at in a separate PR, as they effect the whole application.

Thanks @dated

samharperpittam (approved):

src/router/paths.ts
```diff
@@ -12,6 +12,11 @@ export enum ProfilePaths {
 	ImportWallet = "/profiles/:profileId/wallets/import",
 	ImportWalletLedger = "/profiles/:profileId/wallets/import/ledger",
 	WalletGroupPage = "/profiles/:profileId/network/:networkId",
+	// Messages
+	SignMessage = "/profiles/:profileId/sign-message",
```
goga-m, line 16: @dated is this route along with the `VerifyMessage` route below handled yet? Thinking that if we add them now without handling them it will most probably render an empty screen instead of redirecting to welcome screen :hmhmhm:

dated: good catch! yea, those routes will render an empty page currently, but there aren't link from anywhere yet. i'll comment them out for the sake of correctness asap.

goga-m (approved):

PR #159, opened 2022-08-26
refactor: skip welcome screen if only one profile

alfonsobries:
## Summary

PR https://app.clickup.com/t/2p2urhj

When you only have one profile it available, it shows a "Using profile [profileName]" toast and redirects the user to the deep-linked page

Notice that the "Using profile [profileName]" toast is not part of the task description but used to replace the "You should select a profile to access this URL" used when you have more than one profile.

## Checklist

-   [ ] Ready to be merged


src/domains/profile/pages/Welcome/Welcome.tsx
```diff
@@ -180,6 +180,29 @@ export const Welcome = () => {
...
+		// appears and disappears.
+		let navigateTimeout: ReturnType<typeof setTimeout> | undefined;
+
+		if (!isDeeplink()) {
+			return;
+		}
+
+		navigateTimeout = setTimeout(() => {
+			if (profiles.length === 1) {
+				navigateToProfile(profiles[0]);
```
goga-m, line 195: Hey @alfonsobries I it should be also safe to mark it as clicked here, so it would ignore the user clicks on profile while validating URI.
```suggestion
				isProfileCardClickedOnce.current = true;
				navigateToProfile(profiles[0]);
```

goga-m (approved):

samharperpittam (changes requested):
If a profile is password protected and it's the only profile to exist, when a uri is loaded the 'Password" modal should show by default to log the user in. Currently the user is just taken to the Welcome page.

samharperpittam (changes requested):
Also trying to open a profile that is password protected appears to not work correctly.

- Create password protected profile
- Open URI
- See "Validating URI" toast
- Try to open profile, password modal doens't show.

samharperpittam (approved):

PR #241, opened 2022-11-10
Refactor/delete pending

dated:
## Summary

https://app.clickup.com/t/3221gtc

## Checklist

-   [ ] Ready to be merged


samharperpittam:
The 'Delete' button is now greyed out when viewing the pending transaction with a multisig wallet which makes complete sense. However, I feel we may need a tooltip to explain the reasoning to the user.

Something like....

"Only participants can remove pending transactions"

image

goga-m (approved):

samharperpittam (approved):

dated:
I think the compact styles aren't applied correctly to the delete button :hmhm: it shouldn't have a background color iirc

goga-m:
@dated nice catch!:100: Should be fixed now. Passing in `Button` an `isCompact` prop to remove background only for `danger-icon` variant.

2022-12-02-165238_1203x61_scrot
2022-12-02-165150_1209x51_scrot

PR #251, opened 2022-12-02
refactor: adjust vertical offset for `Modal`

goga-m:
## Summary
Closes https://app.clickup.com/t/320ycgc

## Checklist

-   [x] Ready to be merged


samharperpittam (approved):

dated:
This has the downside that the modal is now very close to the edge of the window

image

goga-m:
@dated @samharperpittam made a few adjustments and the issue with it being too close to the edge should be resolved. 

Instead of changing the vertical padding, I implemented a hook where it sets `overflow-auto` only when window height is less than the actual modal height, while maintaining the same padding as before.

Attaching a video with the behavior:

https://user-images.githubusercontent.com/22020168/205352371-adeca5c9-e60d-48f3-9a04-a26e5f8290db.mp4

Let me know what you think :oleglala:

dated (approved):

PR #286, opened 2023-01-06
feat: migration notifications ui

alfonsobries:
## Summary

Closes part of https://app.clickup.com/t/39xfbkz

**Notes:**

1. This PR hardcodes the notifications using the same format as the `refactor/migration-table` pr, which means it only handles the UI. I think we can handle fetching and storing the notifications in another task (also the action for when notification is clicked). For the moment, I'm typing the migration notification items as `any`
2. This PR also removes the "expanded" version of the regular notifications. That version appears if the users set the "use expanded table" option. [See discussion](https://helloardent.slack.com/archives/C03889W1KHT/p1673280441726099) 
3. Finally, it adjusts the regular notifications to use a small font on mobile devices as per current design (used to use base size)

**To test notifications with a wallet name:**: add wallets `AXzxJ8Ts3dQ2bvBR1tPE7GUee9iSEJb8HX` and/or `AdVSe37niA3uFUPgCgMUH2tMsHF4LpLoiX` that are the ones that are harcoded

## Checklist

-   [ ] Ready to be merged


src/app/components/Notifications/Notifications.contracts.ts
```diff
@@ -24,11 +24,32 @@ export interface NotificationTransactionItemProperties {
...
 }
 
 export interface NotificationsProperties {
 	profile: Contracts.IProfile;
 	onNotificationAction?: (id: string) => void;
 	onTransactionClick?: (item?: DTO.ExtendedConfirmedTransactionData) => void;
 }
+
+export interface NotificationsMigrationItemProperties {
+	// @TODO: assign a proper type for this once defined
```
goga-m, line 36: @alfonsobries if I am not mistaken, I think the type here is `DTO.ExtendedConfirmedTransactionData` :hmhm:

src/app/components/Notifications/Notifications.contracts.ts
```diff
@@ -24,11 +24,32 @@ export interface NotificationTransactionItemProperties {
...
+	transaction: any;
+	profile: Contracts.IProfile;
+	// @TBD (also @TODO set transaction type)
+	onClick?: (transaction: any) => void;
```
goga-m, line 40: @alfonsobries same here. This can be `DTO.ExtendedConfirmedTransactionData` similar to https://github.com/ArdentHQ/arkvault/pull/286/files#diff-493165b65171712d084660d0e622b82a68beb06d2d20c7886ea9a54b9ee0adffR32

samharperpittam (changes requested):
Very minor change, but can we change the "Transactions" header to 18px. Currently it's 16px. Thanks

src/app/components/Notifications/Notifications.contracts.ts
alfonsobries: @goga-m `ExtendedConfirmedTransactionData` contains some equivalences from most of the properties, but it uses methods instead of properties like the fake data I'm using but does not have the polygon `migrationAddress`, The transactions table is also using similar sample data 

I predict that once we save/retrieve the transactions, we will have to define or adapt one of those transaction types both here and in the transaction table. I think it will be a good time to define that type once that happens.

alfonsobries:
@samharperpittam ✅

samharperpittam:
> @samharperpittam ✅

:hmhm:
image

alfonsobries:
@samharperpittam not sure how my changes got lost, but just updated the pr again ✅

samharperpittam (approved):

ItsANameToo (approved):

PR #297, opened 2023-01-11
feat: implement migration transaction table row

goga-m:
## Summary
Closes https://app.clickup.com/t/39xfbmq

For testing,  `DM7UiH4b2rW2Nv11Wu6ToiZi8MJhGCEWhP` wallet already contains a migration transaction.

## Checklist

-   [x] Ready to be merged


samharperpittam (approved):

src/app/components/Label/Label.styles.tsx
```diff
@@ -5,14 +5,15 @@ import { Size } from "@/types";
 
 const baseStyle = tw`inline-block font-semibold overflow-hidden`;
 
-export type ColorType = "primary" | "success" | "danger" | "warning" | "neutral";
+export type ColorType = "primary" | "success" | "danger" | "warning" | "neutral" | "purple";
```
ItsANameToo, line 8: better to give this a generic name instead of the colour. Makes it easier if it changes in the future so you don't end up with `purple` being a blue colour.

Maybe `hint` works, to keep it in line with the variable name as well 🤔

goga-m: Indeed, better to have one consistent name for that variant. Applied the fix @ItsANameToo

ItsANameToo (approved):

PR #310, opened 2023-01-16
refactor: store migration transactions and remove hardcoded 

alfonsobries:
## Summary

Closes https://app.clickup.com/t/865bgdrfu

Note: you may have some migrations already stored in your local storage if you test locally. Since the fake data used is not in the proper format, it will throw an exception. You need to clear storage to test this PR, **Alternatively**

1. Hardcode the following lines inside the `loadMigrations` callback inside the `src/app/contexts/Migration/Migration.tsx` file, just below the  `const migrations = repository!.all();` line (line 97)

```
const newMigrations = migrations.filter((migration) => !migration.id.startsWith("0x"));
repository!.set(newMigrations);
```

2. Open the migration page
3. Remove the lines again since fake migration should have been removed

## Checklist

-   [ ] Ready to be merged


ItsANameToo (approved):

ItsANameToo:
image
this should use a different avatar as it's an ARK address, but that can be resolved later when going over the UI in full

samharperpittam (approved):

dated:
> image
> 
> this should use a different avatar as it's an ARK address, but that can be resolved later when going over the UI in full

just to follow up: the avatar is correct, the address is wrong instead - it should display the address on the polygon network. fixed in https://github.com/ArdentHQ/arkvault/pull/323

PR #328, opened 2023-01-18
feat: migration details modal

alfonsobries:
## Summary

Closes https://app.clickup.com/t/8677418hz

## Checklist

-   [ ] Ready to be merged


samharperpittam (changes requested):
On the `Pending` modal, can we change the description text from `Migration is in the process` to `Migration in progress` please.

image

alfonsobries:
@samharperpittam done ✅

samharperpittam (approved):

src/domains/migration/i18n.ts
```diff
@@ -2,6 +2,17 @@ export const translations = {
 	CONTRACT_PAUSED_MESSAGE:
 		"The migration is temporarily paused. Please check our <linkTwitter>Twitter</linkTwitter> for more details.",
 
+	DETAILS_MODAL: {
+		STEP_PENDING: {
+			DESCRIPTION: "Migration is progress, you will be notified once the process is complete.",
```
ItsANameToo, line 7:
```suggestion
			DESCRIPTION: "Migration in progress, you will be notified once the process is complete.",
```

ItsANameToo (changes requested):
it works, but the modal is slow to open. It seems to be fetching the data after the click, and that can take a while depending on connection speed of the user (or the RPC). It would be better for the UX if we'd open the modal and show a loading state until we have fetched the data so the transition feels smooth and doesn't result in a feeling of the wallet being unresponsive. If we go with a loading state, it will also allow us to show an error in case the migration call fails for some reason. Right now it won't show anything as the catch is ignored

You can throttle the connection in chrome to have a better idea of how it feels on a slower connection.

samharperpittam:
Also when opening a "Success" modal, the modal content can resize, specifically the transaction id truncates. Having a loading state may be beneficial for this too?

alfonsobries:
@ItsANameToo ignore my previous comment; I updated this PR to show a loading skeleton while details are loaded

src/app/contexts/Migration/Migration.tsx
```diff
@@ -88,6 +89,37 @@ export const MigrationProvider = ({ children }: Properties) => {
...
+	const getTransactionStatus = useCallback(
+		async (transaction: DTO.ExtendedConfirmedTransactionData) => {
+			const contractMigrations = await getContractMigrations([`0x${transaction.id()}`]);
+
+			const contractMigration = contractMigrations.find(
+				(contractMigration: ARKMigrationViewStructOutput) =>
+					contractMigration.arkTxHash === `0x${transaction.id()}`,
+			);
+
+			return contractMigration!.recipient === ethers.constants.AddressZero
```
goga-m, line 116: @alfonsobries what will happen if the contractMigration is undefined here? It might stuck on loading state? :hmhm:

ItsANameToo: can be part of https://app.clickup.com/t/86774j27t instead of this PR, we may need handling in more places

ItsANameToo (approved):

PR #335, opened 2023-01-24
refactor: improve migration transaction validation

goga-m:
## Summary
Closes https://app.clickup.com/t/86776g04v

## Checklist

-   [x] Ready to be merged

### DIFF
--- a/src/domains/migration/hooks/use-migration-transactions.ts
+++ b/src/domains/migration/hooks/use-migration-transactions.ts
@@ -1,7 +1,13 @@
 import { useCallback, useEffect, useMemo, useState } from "react";
 import { Contracts, DTO } from "@ardenthq/sdk-profiles";
 import { useConfiguration, useMigrations } from "@/app/contexts";
-import { isValidMigrationTransaction, migrationNetwork, migrationWalletAddress } from "@/utils/polygon-migration";
+import {
+	isValidMigrationTransaction,
+	migrationMinBalance,
+	migrationNetwork,
+	migrationTransactionFee,
+	migrationWalletAddress,
+} from "@/utils/polygon-migration";
 import { Migration } from "@/domains/migration/migration.contracts";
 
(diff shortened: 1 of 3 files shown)


src/domains/migration/hooks/use-migration-transactions.ts
```diff
@@ -49,20 +48,7 @@ export const useMigrationTransactions = ({ profile }: { profile: Contracts.IProf
...
-					return false;
-				}
-
-				if (transaction.recipient() !== migrationWalletAddress()) {
-					return false;
-				}
-
-				return ethers.utils.isAddress(polygonAddress);
-			}),
+		() => latestTransactions.filter((transaction) => isValidMigrationTransaction(transaction)),
```
ItsANameToo, line 51: this filters transactions on the ARKVault side, but won't that cause issues if you have an old valid migration and then a lot of newer invalid migration transactions for an address as it will filter them all out while missing the valid migration on the initial page?

More concrete example:
- address has 101 transactions, the oldest transaction is a valid migration and the 100 other ones are not
- you fetch the newest 100 transactions, all are removed locally because invalid
- there's still 1 transaction left, but it wasn't included in the API call so it's not shown, while that one was the valid transaction

Instead of applying that logic locally, I'd have it apply to the initial query to fetch transactions so you get a response with valid transactions for the overview to avoid scenarios like above 🤔

ItsANameToo (changes requested):

src/domains/migration/hooks/use-migration-transactions.ts
goga-m: @ItsANameToo makes totally sense. Missed the supported query params (`fee.from` & `amount.from`) in the transaction API. I added those now and should be working ok.

Mentioning here also that the only client-side validation that can't be included in the query is the validation of the polygon address (vendorField), so we should now be as close as possible.

Thanks for pointing out :100:

ItsANameToo (approved):

PR #340, opened 2023-01-26
refactor: adjust interval on loading migration transactions

goga-m:
## Summary

https://app.clickup.com/t/86778a87h

* Increases the interval to 5 seconds for loading migrations
* Prevents from loading migrations periodically (rpc calls) when the user is not in migrations page.

## Checklist

-   [x] Ready to be merged


samharperpittam (changes requested):
Migration transaction status doesn't appear to change to "Confirmed" on this PR as discussed on Slack

samharperpittam (dismissed):

src/app/contexts/Migration/Migration.tsx
```diff
@@ -83,9 +84,10 @@ interface Properties {
 
 const MigrationContext = React.createContext<any>(undefined);
 
-const ONE_SECOND = 1000;
+const MIGRATION_LOAD_INTERVAL = 5000;
+const INITIAL_MIGRATION_LOAD_TIME = 1000;
```
ItsANameToo, line 88: shouldn't need this if you make the initial load request the status and only apply an interval afterwards? So you never rely on the interval for the initial check, only when there are pending transaction you add one to occasionally check if the status has changed

samharperpittam (approved):

ItsANameToo (approved):

PR #375, opened 2023-02-12
chore: update dependencies

goga-m:
## Summary

Affected parts to double-check:

* File download (profile export). We could just check if the profile is exported without issues (the file download part, not the profile data)
* Datepicker in transaction export. There was an update on the date picker third-party package we use, we could check if the UI is not affected.
* PWA download. Can just download the PWA from the vercel link and make sure it's loaded ok.

## Checklist

-   [x] Ready to be merged


samharperpittam (changes requested):
When exporting transaction history, after selecting a date range, the modal now provides text which doesn't exist before which is quite ugly. Can we remove this?

Before
image


After
image

goga-m:
@samharperpittam nice one :100: Patched the fix.

ItsANameToo (approved):

samharperpittam (approved):

PR #432, opened 2024-04-09
refactor: handle usernames UI

alfonsobries:
## Summary

Closes https://app.clickup.com/t/86dt6axyj

## Checklist

-   [ ] Ready to be merged


ItsANameToo (changes requested):
some small UI issues

- this seems to cause misalignment for unvotes in the transaction table
image

- when voting for a delegate with no username, it could use additional padding between rank/status to avoid them touching before the dynamic truncation kicks in
image

alfonsobries:
@ItsANameToo ✅

PR #438, opened 2024-04-09
refactor: add username registration and resignation icons

alfonsobries:
## Summary

https://app.clickup.com/t/86dt6ymku

## Checklist

-   [ ] Ready to be merged


src/domains/transaction/components/TransactionDetail/TransactionResponsiveIcon/TransactionResponsiveIcon.tsx
```diff
@@ -24,5 +24,5 @@ export const TransactionResponsiveIcon = ({ icon }: { icon: string }) => {
 	);
 };
 
-export const TransactionDelegateIcon = () => <TransactionResponsiveIcon icon="Delegate" />;
-export const TransactionDelegateResignationIcon = () => <TransactionResponsiveIcon icon="DelegateResignation" />;
+export const TransactionDelegateIcon = () => <TransactionResponsiveIcon icon="UsernameRegistration" />;
```
ItsANameToo, line 27: this changes the delegate icon, that's not what we want as delegate registrations are still used for ARK transactions + this affects sender/recipients when they are a delegate
image

src/domains/transaction/components/UsernameRegistrationDetail/UsernameRegistrationDetail.tsx
```diff
@@ -27,8 +27,6 @@ export const UsernameRegistrationDetail = ({ isOpen, transaction, onClose }: Tra
 		>
 			<TransactionSender address={transaction.sender()} network={transaction.wallet().network()} border={false} />
 
-			{/* @TODO: Use new icons for username registration & resignation types when available.
```
ItsANameToo, line 30: this would need a custom icon instead of the delegate icon it relies on right now

ItsANameToo (changes requested):

alfonsobries:
@ItsANameToo should be ok now

PR #439, opened 2024-04-09
refactor: hide mainsail not supported transactions

alfonsobries:
## Summary

Closes https://app.clickup.com/t/86dt7cxy1

## Checklist

-   [ ] Ready to be merged

### DIFF
--- a/src/domains/transaction/hooks/use-transaction-types.ts
+++ b/src/domains/transaction/hooks/use-transaction-types.ts
@@ -85,10 +85,14 @@ export const useTransactionTypes = ({ wallets = [] }: TransactionTypeProperties 
 		},
 	};
 
+	const mainsailUnsupportedTypes = new Set(["htlcClaim", "htlcLock", "htlcRefund", "ipfs", "secondSignature", "magistrate"]);
+
+	const isMainsail = wallets.some((wallet) => wallet.network().coinName() === "Mainsail");
+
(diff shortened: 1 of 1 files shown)


src/domains/transaction/hooks/use-transaction-types.ts
```diff
@@ -85,10 +85,14 @@ export const useTransactionTypes = ({ wallets = [] }: TransactionTypeProperties
 		},
 	};
 
+	const mainsailUnsupportedTypes = new Set(["htlcClaim", "htlcLock", "htlcRefund", "ipfs", "secondSignature", "magistrate"]);
```
ItsANameToo, line 88: this shouldn't be necessary. The `wallet.transactionTypes()` should instead only list supported types. For example for Mainsail it currently lists these when you log it:
```
["delegateRegistration","usernameRegistration","usernameResignation","delegateResignation","ipfs","multiPayment","multiSignature","secondSignature","transfer","vote"]
```

This still includes IPFS and secondSignature, but that means you need to adjust this on the PSDK side to not return unsupported transaction types

ItsANameToo (changes requested):

ItsANameToo:
closing in favour of https://github.com/ArdentHQ/platform-sdk/pull/65 as that will hide these values in the UI automatically

PR #479, opened 2024-04-24
refactor: username validation rules

shahin-hq:
## Summary
Closes: https://app.clickup.com/t/86dtb91rc

This one is relatively tricky, because it is not bailing out when a validation rule fails, for example, let's say when the given username has a trailing underscore it was showing an invalid error message for few milliseconds and making an API call to validate if username is taken or not, and if it is available the error message was removed which is not ideal, so I had to track the error state, abort the previous request if there is an error there

## Checklist

-   [ ] Ready to be merged


src/domains/transaction/validations/UsernameRegistration.ts
```diff
@@ -1,10 +1,32 @@
 import { Networks } from "@ardenthq/sdk";
 import { ValidateResult } from "react-hook-form";
-import { validatePattern } from "@/utils/validations";
+import { MutableRefObject } from "react";
 import { debounceAsync } from "@/utils/debounce";
 
+const validateUsername = (t: any, value: string): string | undefined => {
```
shahin-hq, line 6: I could have a regex to check all of the cases below but in this case we wouldn't have a descriptive error message, that is why I pick this approach.

ItsANameToo: spaces aren't valid either way, so `trim` shouldn't be needed as it will fail validation on the space itself

ItsANameToo (approved):

PR #485, opened 2024-04-26
[wip] chore: code coverage

alfonsobries:
## Checklist

-   [ ] Ready to be merged

### DIFF
--- a/package.json
+++ b/package.json
@@ -120,7 +120,7 @@
 		"@typescript-eslint/eslint-plugin": "^5.55.0",
 		"@typescript-eslint/parser": "^5.55.0",
 		"@vitejs/plugin-react": "^4.0.0",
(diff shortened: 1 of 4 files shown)


package.json
```diff
@@ -162,7 +162,7 @@
 		"vite-plugin-pwa": "^0.15.1",
 		"vite-plugin-svgr": "^3.2.0",
 		"vite-tsconfig-paths": "^4.0.7",
-		"vitest": "^0.29.8",
+		"vitest": "^0.24.0",
```
alfonsobries, line 165: this is a downgrade, the reason is that it was the only combination I was able to find to make it work, depending on the version used on vitest and @vitest/coverage-istanbul the errors are different but this one seems to work fine and run the tests properly

ItsANameToo (approved):

PR #502, opened 2024-05-09
refactor: prevent using dynamic fee for mainsail

alfonsobries:
## Summary

Closes https://app.clickup.com/t/86dtfmg7p

## Checklist

-   [ ] Ready to be merged


src/app/hooks/use-fees.ts
```diff
@@ -132,7 +133,11 @@ export const useFees = (profile: Contracts.IProfile) => {
 				transactionFees = env.fees().findByType(coin, network, type);
 			}
 
-			if (!!data && (coinInstance.network().feeType() === "size" || type === "multiSignature")) {
+			if (
+				!!data &&
+				!isMainsailNetwork(coinInstance.network()) &&
+				(coinInstance.network().feeType() === "size" || type === "multiSignature")
+			) {
```
ItsANameToo, line 140: can't this be based on the network manifest? feels like there is some logic flaw if we need to explicitly add the mainsail network like this when that network does not have dynamic fees enabled to begin with 🤔

ItsANameToo (changes requested):

ItsANameToo (approved):

PR #503, opened 2024-05-13
fix: handle mainsail deep links

alfonsobries:
## Summary

https://app.clickup.com/t/86dtgpwc0

Requires an update on arkscan https://github.com/ArdentHQ/arkscan/pull/899

Update: now depends on https://github.com/ArdentHQ/platform-sdk/pull/82

## Checklist

-   [ ] Ready to be merged


src/app/hooks/use-search-parameters-validation.tsx
```diff
@@ -245,7 +249,7 @@ export const useSearchParametersValidation = () => {
 
 		const allEnabledNetworks = profileAllEnabledNetworks(profile);
 
-		const coin = parameters.get("coin")?.toUpperCase() || "ARK";
+		const coin = parameters.get("coin") || "ARK";
```
alfonsobries, line 252: I understand the reason for adding `toUpperCase` as a security measure, but that conflicts with the ”Mainsail” coin which is not found if is not in the correct case

An alternative implies modifying `packages/profiles/source/wallet.repository.ts@findByCoinWithNethash` to be case insensitive but IMO is fine to fix it here, think is up to the user to ensure it sends the proper parameter including the correct case

ItsANameToo: if the `Mainsail` coin is the underlying issue, we can adjust it in the psdk network manifest instead since that's only a temporary value (it will be `ARK` eventually). That way we don't introduce a potential breaking change in case someone has a deeplink with a lowercase value that used to work

ItsANameToo (changes requested):

alfonsobries:
@ItsANameToo updated, now it also depends on https://github.com/ArdentHQ/platform-sdk/pull/82

ItsANameToo (approved):

PR #513, opened 2024-05-23
refactor: delegate pagination size per network adjustment

alfonsobries:
## Summary

Closes https://app.clickup.com/t/86dthrkm1

## Checklist

-   [ ] Ready to be merged

### DIFF
--- a/src/domains/vote/components/DelegateTable/DelegateTable.tsx
+++ b/src/domains/vote/components/DelegateTable/DelegateTable.tsx
@@ -11,8 +11,6 @@ import { Pagination } from "@/app/components/Pagination";
 import { EmptyResults } from "@/app/components/EmptyResults";
 import { useBreakpoint } from "@/app/hooks";
 
-const DELEGATES_PER_PAGE = 50;
-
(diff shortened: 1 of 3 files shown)


src/domains/vote/components/DelegateTable/DelegateTable.tsx
```diff
@@ -39,8 +38,9 @@ export const DelegateTable: FC<DelegateTableProperties> = ({
 
 	const columns = useDelegateTableColumns({ isLoading, network: selectedWallet.network() });
 
+	const delegatesPerPage = useMemo(() => (isMainsailNetwork(selectedWallet.network()) ? 53 : 51), [selectedWallet]);
```
ItsANameToo, line 41: this should be taken from the network configuration instead, as that defines the active delegates without having to rely on explicit checks like this

ItsANameToo (changes requested):

ItsANameToo:
also @alfonsobries , better to apply this to `develop` as it affects ARK as well

alfonsobries:
@ItsANameToo updated ✅

PR #537, opened 2024-06-20
feat: implement overwrite modal

shahin-hq:
## Summary
https://app.clickup.com/t/2pkffwd

Scenarios:

- When the form is empty
- When the form is filled and QR lacks some
- When the form is partially filled and QR has all
- When the form is filled and QR lacks some - clear prefilled values unchecked

https://github.com/ArdentHQ/arkvault/assets/132887516/d03e4515-f20a-4fbd-a3cb-69cd6d986fee

## Checklist

-   [ ] Ready to be merged


samharperpittam (dismissed):
The only "issue" I could find was that if the data on the QR code is identical to what is pre-filled we still display the modal like so:

image

But i'm not sure if that's even an issue as such. Will think on it but it looks good 👍

samharperpittam (changes requested):
The modal may need a little love on mobile

1. May have to truncate the Addresses
image

2. When scrolling down, can't read the checkbox cos the button overlay covers it. 

image

shahin-hq:
@samharperpittam please check again:

image

samharperpittam (changes requested):
One final change @shahin-hq  - might need to wrap the memo field as it can be 255 characters long

image

ItsANameToo (approved):

PR #538, opened 2024-06-24
fix: allow musig validator transactions for Mainsail network

shahin-hq:
## Summary
We should let musig wallets to initiate validator registration/resignation transactions.

## Checklist

-   [ ] Ready to be merged


samharperpittam (approved):

src/domains/wallet/pages/WalletDetails/hooks/use-wallet-options.ts
```diff
@@ -96,7 +96,6 @@ const getRegistrationOptions = (wallet: Contracts.IReadWriteWallet, t: TFunction
 
 	if (wallet.balance() > 0 && !wallet.isLedger() && isRestoredAndSynced(wallet)) {
 		if (
-			!isMultiSignature(wallet) &&
```
ItsANameToo, line 99: 🤔

shahin-hq: should be good now :) cc @samharperpittam

samharperpittam (approved):

src/domains/wallet/pages/WalletDetails/hooks/use-wallet-options.ts
```diff
@@ -96,7 +96,7 @@ const getRegistrationOptions = (wallet: Contracts.IReadWriteWallet, t: TFunction
...
+			(isMainsailNetwork(wallet.network()) || !isMultiSignature(wallet)) &&
```
ItsANameToo, line 99: I'll add a card to handle this through psdk, as it would be helpful to have this based purely on the network manifest as that also indicates which musig transactions are supported

ItsANameToo (approved):

PR #558, opened 2024-07-01
chore: update eslint

goga-m:
Closes https://app.clickup.com/t/86dty3n3a 

## Summary

Updates eslint and it's plugins to latest version and migrates configuration to flat config. 
The basic changes in the new eslintconfig.https://github.com/ArdentHQ/arkvault/blob/e9c004e068d1f61ffc329e40075906610d495eda/eslint.config.mjs#L1

Most of the file changes are automatic by eslint fixer here https://github.com/ArdentHQ/arkvault/pull/558/commits/0c0a93f76062a56c5ed5da1ea7ddba8b14b41e11

The main change is that we dropped `eslint-plugin-import` that had rules for the module import handling (ordering, path aliasing et.al) as it doesn't yet support eslint 9 https://github.com/import-js/eslint-plugin-import/issues/2948 and we replaced it with alternative plugins that should do approx the same:
2024-07-03-123252_1087x134_scrot

Created a dedicated task though to monitor & handle the plugin compatibility https://app.clickup.com/t/86du1dtca separately

## Checklist

-   [ ] Ready to be merged


package.json
```diff
@@ -140,12 +146,14 @@
 		"postcss": "^8.4.38",
 		"prettier": "^2.8.4",
 		"prettier-plugin-tailwindcss": "^0.2.5",
+		"prop-types": "^15.8.1",
```
shahin-hq, line 149: Could be deleted

shahin-hq (approved):
good job :100:

PR #578, opened 2024-07-08
chore: update `react-router-dom` package

shahin-hq:
## Summary
Removed `Prompt` as we discusses here: https://helloardent.slack.com/archives/C03889W1KHT/p1720522417368209
tl;dr - to be able to use `Prompt` the new API of react-router should be used which is painful to migrate

## Checklist

-   [ ] Ready to be merged


src/app/components/Link/Link.tsx
```diff
@@ -122,7 +122,8 @@ export const Link = ({
 					{properties.children}
 				</Anchor>
 			) : (
-				<RouterLink component={Anchor} {...properties} />
+				<></>
```
shahin-hq, line 125: Reminder:
https://reactrouter.com/en/main/upgrading/v5#remove-link-component-prop

PR #612, opened 2024-08-14
fix: table rounded borders

patricio0312rev:
# [[general] adjust table rounded borders](https://app.clickup.com/t/86ducpeqz)

## Summary

- Depends on #610 
- Border radius for the header section of the tables has been updated to fix the issue of spacing.
- Snapshots for unit tests have been updated.

image

## Checklist

-   [ ] Ready to be merged


samharperpittam (approved):

tailwind.config.js
```diff
@@ -266,6 +266,7 @@ module.exports = {
 			},
 			borderRadius: {
 				"2.5xl": "1.25rem",
+				table: "0.65rem",
```
ItsANameToo, line 269: this is 10.4 pixels? does it really have to be that or can we go with 10px instead?

ItsANameToo (changes requested):

tailwind.config.js
patricio0312rev: Value fixed ✅

ItsANameToo (approved):

PR #614, opened 2024-08-14
refactor: implement import profile page design changes

shahin-hq:
## Checklist

-   [ ] Ready to be merged

### DIFF
--- a/src/app/assets/svg/import-profile.svg
+++ b/src/app/assets/svg/import-profile.svg
@@ -1,0 +1,17 @@
+<svg viewBox="0 0 24 25" fill="none" xmlns="http://www.w3.org/2000/svg">
+	<g id="Frame 9848">
+		<g id="Group">
+			<g id="money-wallet">
+				<g id="Shape">
+					<path fill-rule="evenodd" clip-rule="evenodd" d="M22.0001 9.5H16.4001C15.6269 9.5 15.0001 8.8732 15.0001 8.1V2.5" fill="#E6EFF9"/>
+					<path d="M22.0001 9.5H16.4001C15.6269 9.5 15.0001 8.8732 15.0001 8.1V2.5" stroke="currentColor" stroke-width="2" stroke-linecap="round" stroke-linejoin="round"/>
+				</g>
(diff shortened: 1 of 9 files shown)


src/domains/profile/pages/CreateProfile/CreateProfile.test.tsx
```diff
@@ -72,22 +62,27 @@ describe("CreateProfile", () => {
...
-		browserAccessMock = vi
-			.spyOn(browserAccess, "fileOpen")
-			.mockResolvedValue(new File([], "picture.png", { type: "image/png" }));
-	});
+	it("should render", async () => {
+		await renderComponent();
 
-	afterEach(() => {
-		browserAccessMock.mockRestore();
+		userEvent.click(screen.getByText("Back"));
```
patricio0312rev, line 68: Not a must, but do you think we can add the `await` before every user event for the tests? It will help to prevent the tests from breaking after we merge with the test refactor branch.

src/app/components/SelectFile/__snapshots__/SelectFile.test.tsx.snap
```diff
@@ -3,55 +3,30 @@
...
         <span
-          class="mr-px hidden font-semibold sm:inline"
+          class="mr-px hidden text-lg font-semibold leading-[21px] sm:inline"
         >
           Drag & Drop or
            
         </span>
         <button
-          class="link ring-focus relative cursor-pointer font-semibold focus:outline-none"
+          class="link ring-focus relative cursor-pointer text-lg font-semibold leading-[21px] focus:outline-none"
```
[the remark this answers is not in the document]
shahin-hq: This version is more readable, I can see the line-height without looking into any config file etc., so I'd prefer to avoid extra abstractions.

src/domains/profile/pages/CreateProfile/CreateProfile.test.tsx
shahin-hq: image

That should be handled when we merge tests.

patricio0312rev (approved):
🔥🔥🔥

samharperpittam (approved):

PR #618, opened 2024-08-16
refactor: implement my votes page design changes

shahin-hq:
## Summary
Closes: https://app.clickup.com/t/86du08j9r

## Checklist

-   [ ] Ready to be merged


.github/workflows/test.yml
```diff
@@ -805,9 +805,9 @@ jobs:
             - name: Test
               uses: nick-invision/retry@v2
               env:
-                  COVERAGE_THRESHOLD_LINES: 98.69
-                  COVERAGE_THRESHOLD_STATEMENTS: 98.75
-                  COVERAGE_THRESHOLD_BRANCHES: 98.48
+                  COVERAGE_THRESHOLD_LINES: 98.67
+                  COVERAGE_THRESHOLD_STATEMENTS: 98.73
+                  COVERAGE_THRESHOLD_BRANCHES: 98.35
```
shahin-hq, line 810: I can't really tell why it was failing due to coverage as it has matching values with a passing test:

Passing test from `develop`: https://github.com/ArdentHQ/arkvault/actions/runs/10486923172/job/29046217452

Failing test: https://github.com/ArdentHQ/arkvault/actions/runs/10486395863/job/29044533608

src/app/components/Table/Table.tsx
```diff
@@ -51,9 +51,14 @@ export const Table = <RowDataType extends Record<never, unknown>>({
...
-	const renderHeaderGroup = (headerGroup: HeaderGroup<RowDataType>) => (
-		<tr {...headerGroup.getHeaderGroupProps()}>{headerGroup.headers.map(renderColumn)}</tr>
-	);
+	const renderHeaderGroup = (headerGroup: HeaderGroup<RowDataType>) => {
+		const { key, ...headerGroupProps } = headerGroup.getHeaderGroupProps();
+		return (
+			<tr key={key} {...headerGroupProps}>
+				{headerGroup.headers.map(renderColumn)}
+			</tr>
+		);
```
shahin-hq, line 60: This is to fix warning:

image

src/domains/vote/components/AddressTable/AddressRow/AddressRowMobile.tsx
```diff
@@ -156,40 +135,23 @@ export const AddressRowMobile = ({ index, maxVotes, wallet, onSelect }: AddressR
...
-									<AddressRowMobileDelegateName name={votes[0].wallet.username()} />
-								</Link>
-							</div>
-						</div>
-					)}
-				</>
-			);
-		}
-
+		// @TODO handle multiple validators
```
shahin-hq, line 138: https://app.clickup.com/t/86duee5xg

.github/workflows/test.yml
ItsANameToo: the reason is that the thresholds depend on a variety of files. Now if you end up refactoring areas in these files, it's possible that the total amount of branches goes down for example. If you don't change the tests (other than removing the ones that are obsolete) it's possible that the coverage goes down because of the amount of branches remaining. Example:
- you have 11 lines of code, 10 are covered = 90.9% coverage
- you remove 1 covered line of code: 10 lines of code remaining, 9 are covered = 90% coverage

this doesn't mean that the code in itself has become less covered, as the exact same assertions are still being made to before, but because it's percentage based it does drop in overall coverage. This will always be a thing; only when we'd have 100% coverage everywhere would this not happen

ItsANameToo (approved):

samharperpittam (approved):

PR #648, opened 2024-09-16
fix: transaction table fixes

patricio0312rev:
# [[wallet details] skeleton states "undefined" when changing tabs](https://app.clickup.com/t/86dunez33)

## Summary

Multiple issues in Transaction Table (desktop view) being addressed here:

- Color for address when an alias is present in dark mode has been adjusted.
image

- Unused prop for `Address` component has been removed and the tests refactored.
- `TransactionTable` now implements `wallet` prop to use the active wallet's currency name.

image

- Unit tests for changes have been updated and refactored.
- Tests for `TransactionRowAddressing` have been renamed correctly after suggestions.

## Checklist

-   [ ] Ready to be merged


src/domains/transaction/components/TransactionTable/TransactionRow/TransactionRowAddressing.test.tsx
```diff
@@ -37,14 +39,14 @@ describe("TransactionRowAddressing", () => {
...
-	it("should render the danger label if transaction is sent", () => {
+	it("should render label with the `to` prefix if transaction is outgoing", () => {
 		const sentFixture = { ...fixture, isSent: () => true };
 		render(<TransactionRowAddressing transaction={sentFixture as any} profile={profile} />);
 
 		expect(screen.getByTestId("TransactionRowAddressing__label")).toHaveTextContent("To");
 	});
 
-	it("should render the success label if transaction is not sent", () => {
+	it("should render label with the `to` prefix if transaction is not outgoing", () => {
```
shahin-hq, line 49: in this case label prefix should be `from` as it is incoming

shahin-hq (approved):

PR #695, opened 2024-10-10
fix: ui adjustments on recipients table

goga-m:
## Summary
Closes https://app.clickup.com/t/86duuft3a

## Checklist

-   [ ] Ready to be merged


samharperpittam (changes requested):
@goga-m  - i've just noticed an issue with the focus state on mobile and was wondering if you could fix it here seeing as you're working on that component anyway.

image

If you look at the right hand side in the corners, you'll see the blue doesn't fill correctly at the edges to the border. 

If it takes longer than 5 minutes to fix then lets move to its own task.

samharperpittam (approved):

ItsANameToo (approved):

PR #703, opened 2024-10-14
fix: correct amount label font size in transaction details

goga-m:
## Summary
Closes https://app.clickup.com/t/86dux5fwt

## Checklist

-   [ ] Ready to be merged

### DIFF
--- a/src/app/components/Notifications/__snapshots__/NotificationsDropdown.test.tsx.snap
+++ b/src/app/components/Notifications/__snapshots__/NotificationsDropdown.test.tsx.snap
@@ -843,7 +843,7 @@ exports[`Notifications > should open and close transaction details modal 1`] = `
                                   class="flex h-full items-center space-x-1"
                                 >
                                   <span
-                                    class="whitespace-nowrap text-sm"
+                                    class="whitespace-nowrap text-sm text-[1rem]"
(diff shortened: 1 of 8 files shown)


src/app/components/Notifications/__snapshots__/NotificationsDropdown.test.tsx.snap
```diff
@@ -843,7 +843,7 @@ exports[`Notifications > should open and close transaction details modal 1`] = `
                                   class="flex h-full items-center space-x-1"
                                 >
                                   <span
-                                    class="whitespace-nowrap text-sm"
+                                    class="whitespace-nowrap text-sm text-[1rem]"
```
ItsANameToo, line 846: it's a snapshot, but I have two issues with it:
- `text-[1rem]` being used instead of `text-base`
- this illustrated that there is an overlap in text classes, e.g. `text-sm text-[1rem]` works now because the 1rem overwrites it, but that is not a given. This will be better to use twmerge with to avoid the duplicate classes

ItsANameToo (changes requested):

ItsANameToo (approved):

samharperpittam (approved):

PR #756, opened 2024-10-22
fix: use available space for  wallet details address in mobile

goga-m:
## Summary
Closes https://app.clickup.com/t/86dux5u8t

## Checklist

-   [ ] Ready to be merged


samharperpittam (changes requested):
Unsure on this @goga-m 

image

I feel that we should leave some padding either side of address.

Can we try truncating the alias and the wallet address on mobile and see how it looks compared to design?

image

samharperpittam (approved):

ItsANameToo (approved):

PR #761, opened 2024-10-23
fix: center profile selection on mobile

shahin-hq:
## Summary
Closes: https://app.clickup.com/t/86dux5r1x

image

## Checklist

-   [ ] Ready to be merged


shahin-hq:
@goga-m the factor that defines where to show profile selection is not determined at that step:
- we show a slider + create profile button if there are profiles created
- we show select profile container if there is a profile - in this case we should center the profile selection part

While skeleton is loading we don't know if there are profiles or not: https://github.com/ArdentHQ/arkvault/blob/e69caffdc7b647f7bea0fe92a4147051e6463aa5/src/app/components/PageSkeleton/PageSkeleton.tsx#L9

if we assume that there will be profiles I can make necessary changes, please let me know.

samharperpittam (approved):

ItsANameToo (approved):

PR #763, opened 2024-10-23
refactor: green background for selected rows

patricio0312rev:
# [[sender/recipient modals] add green background to "selected" address](https://app.clickup.com/t/86duzmkku)

## Summary

- `TableCell` component and styles have been refactored to support `isSelected` props.
- Both the "Select sender" and "Select recipient" tables have been refactored.

## Screenshots

image
image

## Checklist

-   [ ] Ready to be merged

### DIFF
--- a/src/app/components/Table/TableCell.styles.ts
+++ b/src/app/components/Table/TableCell.styles.ts
@@ -1,9 +1,9 @@
-const baseStyle = "flex items-center px-3 my-1 transition-colors duration-100 min-h-11";
+const baseStyle = "flex px-3 items-center my-1 transition-colors duration-100 min-h-11";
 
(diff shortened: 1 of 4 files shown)


samharperpittam (changes requested):
Don't forget dark mode :D 
image

Can use the same styling that we use on the Votes page when a delegate is selected.

patricio0312rev:
@samharperpittam should be fixed now 🌟 
image

patricio0312rev:
@shahin-hq thanks for the review! 🌟 
Actually, I think that the delegate selection page is missing that low padding. As you can see in the designs, the last row always has an extra bottom padding.

image

ItsANameToo (approved):

PR #813, opened 2024-11-13
refactor: set accordions opened by default

patricio0312rev:
# [[portfolio] default to having the list open](https://app.clickup.com/t/86dv5gpgz)

## Summary

- `use-accordion` has been refactored and it will now be opened by default.
- This change will be reflected on wallet accordions (as seen in image), contacts list (in mobile), and custom peers accordion.

image

## Checklist

-   [ ] My changes look good in both light AND dark mode
-   [ ] The change is not hardcoded to a single network, but has multi-asset in mind
-   [ ] I checked my changes for obvious issues, debug statements and commented code
-   [ ] Documentation _(if necessary)_
-   [ ] Tests _(if necessary)_
-   [ ] Ready to be merged

PR #815, opened 2024-11-13
refactor: store accordion preference

patricio0312rev:
# [[portfolio] store table state](https://app.clickup.com/t/86dv5gpje)

## Summary

- Depends on #813 
- `use-accordion` hook has been refactored to store a key for each implementation in local storage. This refactor will allow to store the preference of the user regarding the accordion being opened or not on reload.
- The unit tests have been refactored.

## Checklist

-   [ ] Ready to be merged


samharperpittam (changes requested):
@patricio0312rev - is it possible to make this behaviour function independently of different profiles?

Currently, if you have 2 or more profiles setup, the accordian state is applied to all profiles. For example

- If I have 2 profiles (ProfileA and ProfileB)
- I open profile A
- I set Mainnet accordian to open. Devnet to closed
- This state should just be saved to profile A
- When I open profile B, it should open with the default view.


Currently whatever changes you make to profileA get applied to profileB too.

src/app/hooks/use-accordion.ts
```diff
@@ -1,17 +1,37 @@
...
 
-export const useAccordion = () => {
-	const [isExpanded, setIsExpanded] = useState(true);
+const getStorageKey = (key: string) => `accordion_${key}_isExpanded`;
+
+export const useAccordion = (key: string) => {
+	const storageKey = getStorageKey(key);
+	const [isExpanded, setIsExpanded] = useState<boolean>(() => {
+		const storedValue = localStorage.getItem(storageKey);
+		return storedValue ? JSON.parse(storedValue) : true;
```
shahin-hq, line 9: I think you can remove the `JSON.parse` part, because it stores only bool primitives which don't require deserialization, unless your plan was keeping data in a single key.

image

samharperpittam (approved):

src/app/hooks/use-accordion.ts
shahin-hq: In this case storing 1 or 0 could resolve the issue :eyes:

ItsANameToo (approved):

PR #818, opened 2024-11-14
fix: disable transaction type filter when there are no transactions

shahin-hq:
## Summary
Closes: https://app.clickup.com/t/86dv5gmyt

## Checklist

-   [ ] Ready to be merged


src/domains/transaction/components/Transactions/Transactions.tsx
```diff
@@ -148,6 +149,14 @@ export const Transactions = memo(function Transactions({
 		return !hasEmptyResults || hasFilter || hasMore;
 	}, [activeMode, hasEmptyResults, activeTransactionType, isLoadingTransactions, hasMore]);
 
+	const enableFilter = useMemo(() => {
+		if (isLoadingTransactions || hasFilter) {
```
shahin-hq, line 153: We want to enable filter when transactions are loading or an active filter applied - I mean by that is initially we have all possible options selected this should be understood as no filters applied, when user picks a filter it means there are at least one transaction to display, so we shouldn't disable the element.

samharperpittam (approved):

PR #819, opened 2024-11-14
refactor: dropdown item hover styles

patricio0312rev:
# [[general] add hover states to dropdowns](https://app.clickup.com/t/86dv5gr3t)

## Summary

- Dropdown item hover styles have been updated for all the dropdowns in our dashboard.

https://github.com/user-attachments/assets/69986cdd-6d27-4b6c-9b76-08f6b4b98e0d

## Checklist

-   [ ] Ready to be merged


samharperpittam (approved):

shahin-hq:
@samharperpittam @patricio0312rev this affects all of the dropdowns, including tx type filter, do we want to keep it this way or make it look like in arkscan?

samharperpittam:
> @samharperpittam @patricio0312rev this affects all of the dropdowns, including tx type filter, do we want to keep it this way or make it look like in arkscan?

So for the tx filter (which I think is the only dropdown that contains checkboxes) it would be nice if on hover:
a) the checkbox didn’t have a fill colour
b) the outline of the checkbox changes blue
Then it would largely match Scan

shahin-hq (approved):

PR #872, opened 2024-12-18
fix: set default fee value 

patricio0312rev:
# [[transaction] click on "send all" button crashes the app](https://app.clickup.com/t/86dvfdv8z)

## Summary

- When clicking on the "send all" button in a transaction form, the fee will be taken as zero in case the fees haven't been loaded yet.
- This will prevent the app from crashing since the fees were taken as undefined and making the app to crash.
- The app still validates the amount once the fees have loaded correctly.

https://github.com/user-attachments/assets/6ae0bb36-c102-4f76-8868-759ea925bf9a

## Checklist

-   [ ] Ready to be merged

### DIFF
--- a/src/domains/transaction/components/AddRecipient/AddRecipient.tsx
+++ b/src/domains/transaction/components/AddRecipient/AddRecipient.tsx
@@ -216,7 +216,9 @@ export const AddRecipient: VFC<AddRecipientProperties> = ({
 			return;
 		}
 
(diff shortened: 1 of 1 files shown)


samharperpittam (approved):
Approved as it resolves the issue. 

If a "cleaner" solution is to disable the Send All button until Fees are retrieved then I think that's a perfectly fine solution too.

PR #877, opened 2024-12-19
refactor: send & vote review step redesign

patricio0312rev:
# [[transaction review] adjust Transaction Summary section](https://app.clickup.com/t/86dvf3yxf)

## Summary

- `Send` and `Vote` transactions review steps have been refactored.
- The refactor only considers desktop versions, since the mobile views are still on work.
- The average time is hard coded in this version. Will be updated once the integration with the advanced fees is done.

## Screenshots

- Send (one recipient)
image

- Send (multi payment)
image

- Vote
image

## Checklist

-   [ ] Ready to be merged

### DIFF
--- a/src/domains/transaction/components/TotalAmountBox/TotalAmountBox.tsx
+++ b/src/domains/transaction/components/TotalAmountBox/TotalAmountBox.tsx
@@ -52,7 +52,6 @@ export const TotalAmountBox = ({
 
 	const total = amount + fee;
 	const convertedTotal = convertedAmount && convertedFee ? convertedAmount + convertedFee : undefined;
-
 	return (
 		<DetailWrapper
 			label={t("COMMON.TRANSACTION_SUMMARY")}


samharperpittam (changes requested):
When sending a transaction on a development network, can we hide the fiat value please

image


At least i'm assuming the 0 in the screenshot is the fiat value?

samharperpittam (changes requested):
The height of the confirmation box should be 24px 
image

Currently it's 44
image

samharperpittam (changes requested):
Sorry I should have mentioned this in the task, but for Vote transaction can we remove the "Amount" as that will never be populated for a Vote tx

samharperpittam:
@patricio0312rev - Can we also hide "Amount" on the Transaction Review page for Vote transactions. The "Amount" will always be 0 when voting. 

image

samharperpittam:
https://github.com/ArdentHQ/arkvault/pull/877#pullrequestreview-2517335484

@patricio0312rev  - do you still see this issue?

PR #881, opened 2025-01-06
refactor: remove icons for unused transaction types

goga-m:
## Summary
The transaction type icons are unused now, as well as a few component that were using them (which are also unused after the redesign). This PR just removes them, should not have any changes in UI. 

For testing, we could double-check if the vote info in the wallet page (wallet header) is displaying the icons correctly (vote button & voting info icon in circle). Other than that, no other UI should be affected.

Closes https://app.clickup.com/t/86dvc4w7p 

## Checklist

-   [ ] Ready to be merged


src/domains/transaction/components/TransactionTable/TransactionRow/TransactionRowMode.test.tsx
shahin-hq, line 1: We could remove its contracts as well: https://github.com/ArdentHQ/arkvault/blob/c7f12e38a357c8caf12134e462a10a3c766a990d/src/domains/transaction/components/TransactionTable/TransactionRow/TransactionRowMode.contracts.ts#L4

ItsANameToo (approved):

shahin-hq (approved):

PR #900, opened 2025-01-20
feat: implement addresses side panel

shahin-hq:
## Checklist

-   [ ] Ready to be merged


package.json
```diff
@@ -95,6 +95,7 @@
 		"string-hash": "^1.1.3",
 		"tailwind-merge": "^2.5.2",
 		"tippy.js": "^6.3.7",
+		"usehooks-ts": "^3.1.0",
```
shahin-hq, line 98: This package includes handful amount of useful hooks and it is tree shakeable, in my case I am using `useLocalStorage` which helps to store onboarding messages state.

src/app/components/Checkbox/Checkbox.tsx
```diff
@@ -8,25 +8,32 @@ type CheckboxProperties = {
...
-					color === "danger",
-				"checked:text-theme-hint-500 [&:not(:checked)]:hover:border-theme-hint-500": color === "hint",
-				"checked:text-theme-primary-600 [&:not(:checked)]:hover:border-theme-primary-600": color === "info",
-				"checked:text-theme-primary-600 checked:hover:text-theme-primary-700 [&:not(:checked)]:hover:border-theme-primary-600":
-					color === "success",
-				"checked:text-theme-warning-600 [&:not(:checked)]:hover:border-theme-warning-600": color === "warning",
-			}),
-			props.className,
-		)}
-		{...props}
```
shahin-hq, line 28: This was overriding `className` prop when it exists

src/domains/setting/components/PasswordRemovalConfirmModal/__snapshots__/PasswordRemovalConfirmModal.test.tsx.snap
```diff
@@ -172,7 +172,7 @@ exports[`PasswordRemovalConfirmModal > should render 1`] = `
               data-testid="Form"
             >
               <fieldset
-                class="[&>.FormLabel]:text-theme-primary-600 focus-within:[&>.FormLabel]:text-theme-primary-600 flex min-w-0 flex-col"
+                class="[&:focus-within_.FormLabel]:text-theme-primary-600 dark:[&:focus-within_.FormLabel]:text-theme-primary-500 flex min-w-0 flex-col"
```
shahin-hq, line 175: That wasn't necessarily result of my changes but changes from disabled tests. Snapshot action added these.

samharperpittam (changes requested):
@shahin-hq - Overall really good, just a couple of very minor thing I noticed

1. The corners of the onbosarding tooltip don't appear to be rounded enough when compared with the designs.

2. Onboarding tooltip should only not be shown again if the "Got It" button is clicked. Currently if you open and then close the side panel it won't show again. I think making sure it's binded to the "Got It" button ensures that the message has been read.

3. When searching for an address that can't be found, the "Select All" checkbox is automatically checked. I'm inclined to think it would be better behaviour if the checkbox was disabled if no search results are returned.

4. Lastly this is more of a question that i'm thinking on and additional opinions are welcome, but when searching for an address, if the user closes the side panel and then opens again, should the search term persist? Currently it does. I can see pro's and cons both ways on this one but I think i'm leaning towards the search should be cleared when the panel is closed. Open to opinions though.

samharperpittam (approved):

PR #917, opened 2025-01-27
refactor: use wallet alias hook

patricio0312rev:
# [[general] handle usernames in UI](https://app.clickup.com/t/86dvmaf3k)

## Summary

- `TransactionRowAddressing` has been updated, it was previously using the sender address to get the alias, it now uses the recipient one.
- The `use-wallet-alias` hook has been refactored, following this logic:
  - First, it will look for the `displayName` of the wallet in the profile.
  - Secondly, it searches for the name of the address in the `contact list`.
  - Then, it goes for the address and its `on-chain name`.
  - If it has none of them, it returns an `undefined` value in the alias and will simply render the address in case of components such as `TransactionRowAddressing`.
- Unit tests have been updated to give coverage to these changes.

image

## Checklist

-   [ ] Ready to be merged


src/app/hooks/use-wallet-alias.tsx
```diff
@@ -52,19 +53,11 @@ const useWalletAlias = (): HookResult => {
...
-						alias = delegateUsername;
-					}
-
 					return {
 						address,
-						alias,
+						alias: wallet.displayName(),
 						isContact: false,
-						isDelegate: !!delegateUsername,
+						isDelegate: !!getDelegateUsername(network),
```
shahin-hq, line 60: :thinking: I don't think we should be using delegate username or etc, because username is no longer related to being delegate, it is just another property that wallet may have. So, I believe all the logic related to delegate should be gone, but let's clarify it with @ItsANameToo.

ItsANameToo: the property is wrong with this logic though, as it assumes `isDelegate: false` for an address without a username, while that address can still be a delegate (validator).

What you'd want to do here is:
- adjust `delegate` name to `validator` instead so it matches Mainsail (here and in other checks that use the `delegate` word)
- adjust the logic to properly define `isValidator` or omit it completely if it's no longer necessary (and refactor the corresponding components that made use of it)

ItsANameToo (changes requested):

src/app/hooks/use-wallet-alias.tsx
shahin-hq: Right, but even in this case `username` should not be used to determine if address is a delegate or not:

```javascript
						const delegate = env.delegates().findByAddress(network.coin(), network.id(), address);
						return delegate.username();
```

And regarding the `isDelegate` prop, I think we should be checking its usages as well because wallet alias is no longer depend on `isDelegate`.

src/app/hooks/use-wallet-alias.test.tsx
```diff
@@ -79,79 +53,64 @@ describe("useWalletAlias", () => {
...
-			result.current.getWalletAlias({
-				address: wallet.address(),
-				network: wallet.network(),
-				profile,
-			}),
+			result.current.getWalletAlias({ address: wallet.address(), profile, username: "delegate_username" }),
 		).toStrictEqual({
 			address: wallet.address(),
-			alias: "delegate username",
+			alias: "delegate_username",
```
shahin-hq, line 112: Minor change: `xyz_username` over `delegate_username`, as delegate is misleading.

shahin-hq:
Overall look good @patricio0312rev, good job, left a question.

shahin-hq (approved):

samharperpittam (approved):

PR #924, opened 2025-01-29
refactor: handle from/to addresses in multiview (mobile views)

patricio0312rev:
# [[portfolio] handle from/to addresses in multi-view](https://app.clickup.com/t/86dvuxd53)

## Summary

- This first part is only focused on 320px, 640px and 768px views.
- The `TransactionRow` and `TransactionRowMobile` components have been refactored to support the `hideSender` prop, which activates/deactivates the advanced mode of the `addressing` column.
- `TransactionRowAddressing` has been refactored to support `variant` and `isAdvanced` props. When the user has selected more than 1 wallet, the advanced mode of the addressing column activates and two of these components will render, the first always for the `sender` and the second one always for the `recipient`.
- `TransactionRowAddressing` has been refactored and subdivided into child components for better readability and reduce the complexity when working with this component.
- Unit tests have been updated to support these changes.
- The transaction table's `Value` header has been renamed `Amount` to match the designs.

## Screenshots

- 320px
image

- 640px
image

- 768px
image

## Checklist

-   [ ] Ready to be merged

### DIFF
--- a/src/app/components/Input/Input.tsx
+++ b/src/app/components/Input/Input.tsx
@@ -117,6 +117,7 @@ export const Input = React.forwardRef<InputElement, InputProperties>(
 			style,
 			suggestion,
 			value,
+			readOnly,
 			...properties
 		}: InputProperties,
 		reference,
@@ -132,7 +133,7 @@ export const Input = React.forwardRef<InputElement, InputProperties>(
 
 		const focusReference = useRef<InputElement>(null);
 
-		reference = isFocused ? focusReference : reference;
+		const inputReference = isFocused ? focusReference : reference;
 
 		useEffect(() => {
 			if (isFocused && focusReference.current) {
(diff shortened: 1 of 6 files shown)


samharperpittam (changes requested):
When viewing 640, there appears to be certain scenarios where the "To" and "From" badges are different sizes. I _think_ it happens when a sender or receive does not have an address name but not 100% on that.

image

See image above.
From = 44.8x21
To = 48.21

Another example
image

samharperpittam (changes requested):
The `Value (ARK)` text should read `Amount (ARK)`

Design
image

samharperpittam (changes requested):
@patricio0312rev 

image

These all read `Value (ARK)` now instead of `Amount (ARK)`. Can you check if it's the same for you please?

samharperpittam (changes requested):
@patricio0312rev  - last change request from me. In Multiwallet view, the "Amount" value shouldn't have the green/yellow/grey background and the text should be black.

Design
image

Implementation
image

samharperpittam (approved):

PR #933, opened 2025-01-31
refactor: display username in transaction details modal

patricio0312rev:
# [[tx details] usernames not considered](https://app.clickup.com/t/86dvvqr13)

## Summary

- `useTransactionRecipients` hook has been refactored to now get the username when using `getWalletAlias` by following the logic of previous tickets.

image

## Checklist

-   [ ] Ready to be merged

### DIFF
--- a/src/domains/transaction/components/TransactionTable/TransactionRow/TransactionRow.test.tsx
+++ b/src/domains/transaction/components/TransactionTable/TransactionRow/TransactionRow.test.tsx
@@ -87,6 +87,31 @@ describe("TransactionRow", () => {
 		expect(asFragment()).toMatchSnapshot();
 	});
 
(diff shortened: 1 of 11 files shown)


samharperpittam (approved):

ItsANameToo:
@patricio0312rev @goga-m it would be better if that line does not get hidden and show the address instead if no username is fetched yet

ItsANameToo (changes requested):
on this PR, the recipient is no longer using the same truncation as the sender:
image

compared to feat/evm:
image

I also think that we should consider some improvements to avoid needless network requests, e.g.:
- if sender === recipient, don't fetch recipient as we already have the details for it
- preferably we also cache these recipient calls so subsequent transaction details don't require network requests for the same data, but that can be a separate card and PR to improve it

PR #935, opened 2025-02-02
refactor: make import dropdown selectable only

patricio0312rev:
# [[import] make import dropdown selectable only](https://app.clickup.com/t/86dvw0ed7)

## Summary

- `Input` has been refactored to support `readOnly` prop correctly, since it was previously ignoring it.
- Unit tests for Import Wallet have been updated.
- Snapshots has been updated for this change
- Implementations of `Select` component in settings page have been fixed

https://github.com/user-attachments/assets/537fa778-41fa-4457-a171-9756403794b1

## Checklist

-   [ ] Ready to be merged


ItsANameToo (changes requested):
this crashes the component when it's typed into. You can try it by going to settings > general and typing in one of the dropdowns there
image

samharperpittam (approved):
Approving as the original issue described in the task is fixed. 

However, looking at other areas of Vault (such as Settings and Contacts), i'm thinking we should probably implement the same logic globally for all dropdowns. But that can be a separate task if required.

ItsANameToo (approved):

PR #938, opened 2025-02-03
refactor: total amount label styles in multi-address tx table

patricio0312rev:
# [[portfolio] adjust value labels in transactions table](https://app.clickup.com/t/86dvvnk40)

## Summary

- Depends on #924
- `TransactionTotalLabel` has been refactored to support a variant with no label styles.
- This variant is now being activated in both the desktop and mobile variants of the `TransactionTableRow` when `hideSender` is `false`

image

## Checklist

-   [ ] Ready to be merged


samharperpittam:
> @samharperpittam oki, I've updated that, but the logic to update the column name and the change you requested are in #940
> 
> This PR is just for the label styles 🌟

@patricio0312rev  - sorry i think there's been some confusion with my latest comment.

The column header should retain the `(ARK)` text. However, the actual value in the table doesn't require the ARK text as it's stated in the header

image

samharperpittam (approved):

ItsANameToo (approved):

PR #941, opened 2025-02-03
refactor: validator resignation page to support multiple wallets

shahin-hq:
## Checklist

-   [ ] Ready to be merged


samharperpittam (approved):
Approved but think we will need a follow up task to handle addresses that haven't previously registered as a validator. As currently all addresses can be selected even if it's not possible to resign a validator. Will discuss this issue and get a design created.

PR #944, opened 2025-02-04
fix: prevent unnecesary re-renders in use portfolio hook

patricio0312rev:
# [[vault] Error: maximum update depth exceeded](https://app.clickup.com/t/86dvv8cf5)

## Summary

- A new check has been added in `usePortfolio` hook to prevent unnecessary updates for `setConfiguration`, which caused this issue when having no wallets imported in a profile.

## Checklist

-   [ ] Ready to be merged

### DIFF
--- a/src/domains/transaction/pages/SendValidatorResignation/FormStep.tsx
+++ b/src/domains/transaction/pages/SendValidatorResignation/FormStep.tsx
@@ -5,19 +5,37 @@ import { useTranslation } from "react-i18next";
 import { Alert } from "@/app/components/Alert";
 import { FormField, FormLabel } from "@/app/components/Form";
 import { FeeField } from "@/domains/transaction/components/FeeField";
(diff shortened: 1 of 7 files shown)


src/domains/portfolio/hooks/use-portfolio.ts
```diff
@@ -121,7 +121,7 @@ export const usePortfolio = ({ profile }: { profile: Contracts.IProfile }) => {
 	const allAddresses = addresses.all();
 
 	useEffect(() => {
-		if (selectedAddresses.length === 0) {
+		if (selectedAddresses.length === 0 && allAddresses.length > 0 && !isEqual(selectedAddresses, allAddresses)) {
```
ItsANameToo, line 124:
```suggestion
		if (selectedAddresses.length === 0 && !isEqual(selectedAddresses, allAddresses)) {
```
this would have the same effect right?

shahin-hq (approved):

PR #951, opened 2025-02-04
refactor: display username in transaction details modal

patricio0312rev:
# [[tx details] usernames not considered](https://app.clickup.com/t/86dvvqr13)

## Summary

- `useTransactionRecipients` hook has been refactored to now get the username when using `getWalletAlias` by following the logic of previous tickets.
- `Address` component reference now calculates the width to truncate considering the alias when it updates.

image

## Checklist

-   [ ] Ready to be merged


samharperpittam (changes requested):
There seems to be a couple of issues with this PR

Here's the setup I have

- Create new profile
- Import the following address `0x6F0182a0cc707b055322CcF6d4CB6a5Aff1aEb22`
- This address has a username registration of `php_sdk_tester`

When opening various transaction detail modals, I'm expecting any interaction from the address `0x6F0182a0cc707b055322CcF6d4CB6a5Aff1aEb22` to contain the name `php_sdk_tester`. However, there's a couple of issues.

1. Transaction to Contract
When viewing the transaction details for a transaction to Contract, it appears the Contract address is incorrectly given the username `php_sdk_tester`. Not `0x6F0182a0cc707b055322CcF6d4CB6a5Aff1aEb22`

image

---

General Transactions

When viewing transactions that aren't to contract, it appears that the `Address Name` is used rather than the `Username`. In the image below, the recipient should be `php_sdk_tester`.

image

samharperpittam (approved):

src/domains/transaction/components/TransactionDetailModal/TransactionDetailModal.tsx
```diff
@@ -72,7 +72,7 @@ export const TransactionDetailContent = ({
 						network={transaction.wallet().network()}
 						recipients={recipients.map(({ address, alias }) => ({
 							address,
-							alias,
+							alias: isContractTransaction(transaction) ? undefined : alias,
```
ItsANameToo, line 75: this just sets any contract to have no alias, so if you were to add a contract as a contact in the wallet it wouldn't show up either 🤔 why was it showing the wrong alias in the first place? shouldn't the alias be based on the recipient and therefore show the right value irregardless of it being a contract or regular address?

ItsANameToo (changes requested):

ItsANameToo (approved):

PR #956, opened 2025-02-06
feat: implement network switch

goga-m:
## Summary

Closes https://app.clickup.com/t/86dvv87bz

* Addition of mainsail mainnet network, with the default configs from PSDK.
* `usePortfolio` hook now mutates & filters wallets based on the active network
* A new hook `useActiveNetwork` is implemented, to provide the selected network, which can be used throughout the app and it's decoupled from wallets.
* Network Step is removed from wallet creation & wallet import flows. It should always use the active network from the navigation. 

### For testing
1. Enable Testnet toggle in  Settings -> General -> Show Development Network.
2. Import, create wallets and see them persisted per active network.
3. The address selection is also per network.
4. Votes page should also render data filtered by the active network.
5. All the pages are reactive to the navigation network toggle,vso changing the network should re-render the page if necessary based on the selection.

A few more notes:
* Existing profile wallet selections may not work, as the underlying structure has been changed and we need to re-apply the selection for existing profiles.
* Wallet creation and import should work without issues with mainnet. However, as the existing config uses endpoints from testnet, you might see transactions & other network data from testnet. They can be ignored, and we can retest once we have the mainnet ready and pass the correct config in PSDK.
* Default network is Testnet.
* Mobile designs are not added yet, as we can inject the toggle once the navigation UI refactor will be ready in https://github.com/ArdentHQ/arkvault/pull/958
UPDATE:  UI designs should be updated as #958 is merged.

## Checklist

-   [ ] Ready to be merged


samharperpittam (changes requested):
@goga-m  - noticed a couple of issues with this. 

1. Default network not restored after disabling developer network setting

- Create Profile
- Import Address on Testnet (default network)
- Settings -> Enable developer network
- Change network to Mainnet
- Import Address (give address custom name of Mainnet)
- Settings -> Disable Developer Networks
- Network doesn't restore to default network (Testnet). The address "Mainnet" is still in view.


---

2. When creating/importing an address for the "Default" network, the address should also be added to the profile for the secondary network (mainnet in this case). Currently, this does not happen. 

For example, what _should_ happen is:
- Create Profile
- Import 1 address on default network (Testnet)
- Enable Developer Network in Settings
- Change network toggle to secondary network (Mainnet)
- The address that was imported on the default network should also exist on the secondary network

Instead when switching network, the user is asked to either Create or Import an address.

samharperpittam (changes requested):
image

When the "Show Development Network" network toggle is disabled in Settings, the network toggle within the dropdown is still visible when it should be hidden.

samharperpittam (approved):

ItsANameToo:
got some workflows to look at @goga-m 👀

ItsANameToo (approved):

PR #971, opened 2025-02-13
refactor: transaction amount column rename

patricio0312rev:
# [[portfolio] table in single address view incorrectly displays "Value"](https://app.clickup.com/t/86dvyzn3x)

## Summary

- `Value` column has been renamed to simply `Amount` and the logic behind toggling this title name has been removed.
- Mobile transaction row has also been refactored to support this change.
- Unit tests have been refactored to support these changes.

image

## Checklist

-   [ ] Ready to be merged

### DIFF
--- a/src/domains/transaction/components/TransactionTable/TransactionRow/TransactionRow.test.tsx
+++ b/src/domains/transaction/components/TransactionTable/TransactionRow/TransactionRow.test.tsx
@@ -131,7 +131,7 @@ describe("TransactionRow", () => {
 									isLedger: () => false,
 									network: () => ({
 										coin: () => "DARK",
-										id: () => "ark.mainnet",
+										id: () => "ark.devnet",
 									}),
 								}),
 							} as any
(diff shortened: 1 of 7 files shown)


src/domains/transaction/components/TransactionTable/TransactionRow/TransactionRow.test.tsx
```diff
@@ -15,6 +15,10 @@ describe("TransactionRow", () => {
 		wallet: () => ({
 			...TransactionFixture.wallet(),
 			currency: () => "DARK",
+			network: () => ({
+				coin: () => "DARK",
+				id: () => "ark.mainnet",
```
ItsANameToo, line 20: although it will work, for consistency I'd prefer to use the right `id` when using it in combination with DARK. Applies to the other changes in this PR as well

```suggestion
			network: () => ({
				coin: () => "DARK",
				id: () => "ark.devnet",
```

ItsANameToo (changes requested):

src/domains/transaction/components/TransactionTable/TransactionRow/TransactionRow.test.tsx
ItsANameToo: @patricio0312rev keywords: **Applies to the other changes in this PR as well**

ItsANameToo (approved):

PR #986, opened 2025-02-18
refactor: prefill sender when having only one wallet

patricio0312rev:
# [[general] prefill sender if only 1 address exists](https://app.clickup.com/t/86dw0mn87)

## Summary

- The sender will now prefill when having only one wallet and we access the `SendTransfer` form

## Checklist

-   [ ] Ready to be merged


src/app/components/NavigationBar/NavigationBar.blocks.tsx
```diff
@@ -304,7 +304,10 @@ export const NavigationBarFull: React.FC<NavigationBarFullProperties> = ({
 	const handleCloseReceiveFunds = useCallback(() => setSelectedWallet(undefined), [setSelectedWallet]);
 
 	const sendButtonClickHandler = useCallback(() => {
-		const sendTransferPath = `/profiles/${profile.id()}/send-transfer`;
+		let sendTransferPath = `/profiles/${profile.id()}/send-transfer`;
+		if (wallets.length === 1) {
+			sendTransferPath = `/profiles/${profile.id()}/wallets/${wallets[0].id()}/send-transfer`;
+		}
```
ItsANameToo, line 310: would it be possible to have the address selection dropdown default to the first entry if there is only 1 option instead? Then we don't have to make these route adjustments in places where we link to it as the form itself will take care of it and make it less error prone 🤔

ItsANameToo (approved):

PR #994, opened 2025-02-19
refactor: move import address to side panel

shahin-hq:
Closes: https://app.clickup.com/t/86dvv8pxw
Design: https://www.figma.com/design/OX18z9oCfYiV4cW35bzxsu/ARKVault-Mainsail?node-id=23574-69801&m=dev

## Summary
This PR moves import address flow to side panel. I had move files to `portfolio` domain, that is why there are too many files _changed_. I also had to add icons and adjust flow for an extra step, the review process should be pretty easy. 

[Screencast from 2025-02-20 20-26-20.webm](https://github.com/user-attachments/assets/44a78a0a-1e7b-440a-ba1e-4719baaf67cc)

## Checklist

-   [ ] Ready to be merged


.github/workflows/test.yml
```diff
@@ -857,10 +857,10 @@ jobs:
             - name: Test
               uses: nick-invision/retry@v2
               env:
-                  COVERAGE_THRESHOLD_LINES: 94.78
-                  COVERAGE_THRESHOLD_FUNCTIONS: 97.22
-                  COVERAGE_THRESHOLD_STATEMENTS: 94.9
-                  COVERAGE_THRESHOLD_BRANCHES: 91.73
+                  COVERAGE_THRESHOLD_LINES: 89.82
```
shahin-hq, line 860: I moved stuff from `wallet` domain, that is why adjusted coverage

shahin-hq:
@goga-m thanks, should be good now :eyes:

samharperpittam (changes requested):
When Importing via Public Key or Secret, the "Import Complete" page doesn't get displayed. 
image

shahin-hq:
@samharperpittam that behavior is not specific for the public key or secret, you can see the same behavior when there are no addresses in the profile. When we import an address the profile triggers logic to show dashboard, hide import options which also caused side panel to unmount. To fix that I lifted side panel state to `Portfolio` page, it should work as expected now.

samharperpittam (approved):

ItsANameToo (approved):

PR #1021, opened 2025-03-04
refactor: sidepanel blur effect

patricio0312rev:
# [[general] blur background more when side panel is open](https://app.clickup.com/t/86dw480pn)

## Summary

- Backdrop blur of 24px has been added to the overlay container in side panels to match the designs.

image

image

## Checklist

-   [ ] Ready to be merged


samharperpittam (commented):
I quite like this blur effect

There is an oddity though where the onboarding tooltip gets displayed through the blur but i'm happy to fix that in a different task to keep things simple

image

PR #1024, opened 2025-03-04
fix: fee input field focus

patricio0312rev:
# [[advanced fee] field focus issue](https://app.clickup.com/t/86dw4vjxg)

## Summary

- In `FeeField` component, `watch()` has been removed from watching `gasPrice` and `gasLimit` to prevent excessive re-renders.
- Instead, added `useRef` to allow state persistence without causing rerenders that were removing the focus from the inputs.
- Also, memoization has been implemented to prevent unnecessary re-renders.

https://github.com/user-attachments/assets/88b4cb58-773f-4bd8-b4a3-fc578a6c6342

## Checklist

-   [ ] Ready to be merged


samharperpittam (changes requested):
@patricio0312rev  - when entering a number in one of the Advanced Fee fields, the the field seems to automatically lose its focus state after a second or so, which I find a bit odd to use. Is there a particular reason for that?

ItsANameToo:
you still lose focus after typing, it's just delayed to 300ms now

PR #1028, opened 2025-03-05
fix: wallet alias usage

patricio0312rev:
# [[general] addresses with 0 balance no longer display a name](https://app.clickup.com/t/86dw57yma)

## Summary

- The issue of the ticket started due to an error related to the wallets not being synced:
```
Error: This wallet has not been synchronized yet. Please call [synchroniser().identity()] before using it.
```
- To solve this, this PR is setting the local name as the alias initial value once it has been fetched (and if it exists), and then continues relying on the following logic selected by the user in the settings in case its wallet has transactions on chain.
- In addition, wallet alias will now display in transaction details modal (it also stopped appearing)

## Checklist

-   [ ] Ready to be merged


ItsANameToo (changes requested):
does not seem to work as the sidebar still lacks address names

samharperpittam (approved):

ItsANameToo (approved):

PR #1030, opened 2025-03-06
fix: disabled button issue in create address flow

shahin-hq:
## Summary
Closes: https://app.clickup.com/t/86dw0pev4

[Screencast from 2025-03-06 13-57-17.webm](https://github.com/user-attachments/assets/67e41991-bd24-4fcf-a1b0-ac78fd9068b3)

## Checklist

-   [ ] Ready to be merged


samharperpittam (approved):
PR works for Create flow. However I also tested the Import flow and the same issue exists there. Task: https://app.clickup.com/t/86dw5xcwe

ItsANameToo (approved):

.github/workflows/test.yml
```diff
@@ -857,10 +857,10 @@ jobs:
             - name: Test
               uses: nick-invision/retry@v2
               env:
-                  COVERAGE_THRESHOLD_LINES: 89.82
+                  COVERAGE_THRESHOLD_LINES: 89.8
```
shahin-hq, line 860: Added unit test to cover my  changes, however, it is still complaining probably because the addition I made is in `portfolio` domain while deletion is in `domain` domain.

PR #1036, opened 2025-03-07
fix: selected addresses in side panel

shahin-hq:
## Summary
Closes: https://app.clickup.com/t/86dw5xa92 & https://app.clickup.com/t/86dw3g6da

This PR fixes selected addresses in the side panel after import & create

[Screencast from 2025-03-07 19-34-19.webm](https://github.com/user-attachments/assets/1563c714-2679-4a05-9038-b85b387ef1d8)

## Checklist

-   [ ] Ready to be merged

### DIFF
--- a/src/domains/portfolio/components/PortfolioHeader/PortfolioHeader.tsx
+++ b/src/domains/portfolio/components/PortfolioHeader/PortfolioHeader.tsx
@@ -374,6 +374,7 @@ export const PortfolioHeader = ({
 				onClose={(addresses) => {
 					setSelectedAddresses(addresses);
 				}}
+				key={selectedAddresses.join('-')}
 				open={showAddressesPanel}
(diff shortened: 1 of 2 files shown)


src/domains/portfolio/components/PortfolioHeader/PortfolioHeader.tsx
```diff
@@ -374,6 +374,7 @@ export const PortfolioHeader = ({
 				onClose={(addresses) => {
 					setSelectedAddresses(addresses);
 				}}
+				key={selectedAddresses.join('-')}
```
shahin-hq, line 377: We are passing `selectedAddresses` as a prop to the `AddressesSidePanel` which uses it to form a state, so we need to enforce it clear inner state when selected addresses change, that is when `key` is useful.

ItsANameToo (approved):

PR #1070, opened 2025-03-17
refactor: wallet encryption toggle

patricio0312rev:
# [[wallet encryption] implement updated design](https://app.clickup.com/t/86dw75w9m)

## Summary

- Wallet encryption toggle has been redesigned. A new `WalletEncryptionBanner` component has been created containing the new design, replacing both implementations in the `Create` and `Import` wallet flows.
- With this new toggle, a new `Accept Responsibility` checkbox has been introduced and registered into the forms, which is now required to continuing registered the wallets.
- Dark mode designs have also been updated for this component.
- Unit tests and translation files have been updated.

## Screenshots

image
image

## Checklist

-   [ ] Ready to be merged

### DIFF
--- a/src/domains/portfolio/components/CreateWallet/ConfirmPassphraseStep.tsx
+++ b/src/domains/portfolio/components/CreateWallet/ConfirmPassphraseStep.tsx
@@ -4,7 +4,7 @@ import { useTranslation } from "react-i18next";
 
 import { MnemonicVerification } from "@/domains/wallet/components/MnemonicVerification";
 import { Checkbox } from "@/app/components/Checkbox";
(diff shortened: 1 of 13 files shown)


samharperpittam (changes requested):
@patricio0312rev  - when enabling the toggle, after the component expands the text seems to jump around a bit, and the dotted line seems to shrink which looks a bit odd. Can anything be done to prevent that from happening?


https://github.com/user-attachments/assets/8c43e2a9-efd3-4c12-b5bd-5887cdd7cdec



It would be nice if it could keep the transition but without the text moving about.

samharperpittam (approved):

ItsANameToo (approved):

PR #1082, opened 2025-03-19
fix: switching profiles bug

alfonsobries:
## Summary

Modifies the `use-profile-synchronizer` hook to handle status by profile. This is a breaking change, so I recommend comprehensive testing

https://app.clickup.com/t/86dw8yz70

## Checklist

-   [ ] Ready to be merged

### DIFF
--- a/src/app/App.blocks.tsx
+++ b/src/app/App.blocks.tsx
@@ -5,7 +5,7 @@ import { useErrorBoundary } from "react-error-boundary";
 import { ToastContainer } from "react-toastify";
 
 import { ConfirmationModal } from "@/app/components/ConfirmationModal";
-import { useEnvironmentContext, useNavigationContext } from "@/app/contexts";
+import { useConfiguration, useEnvironmentContext, useNavigationContext } from "@/app/contexts";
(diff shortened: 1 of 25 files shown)


samharperpittam (changes requested):
@alfonsobries - is there anything in this PR that may have an impact on nonce value retrieval? 

I've managed to get an address of mine in a state where it can no longer send transaction due to incorrect nonce used

image

I cleared all localstortage and created a new profile and was able to replciate it again. Seemingly by just sending a bunch of transactions on repeat but i'm unable to reliably reproduce with a specific scenario but will keep trying. 

I then exported this profile and imported it to `feat/evm` and Vault was able to send transactions without issue as it retrieved the nonce correctly.

alfonsobries:
@samharperpittam I wasn’t able to replicate it, I suspect it was a coincidence. It sounds like some kind of cache issue or something like that where it’s not properly getting the last used nonce, still there’s nothing in this PR that I can think of that would affect it

samharperpittam (approved):

ItsANameToo (approved):

PR #1095, opened 2025-03-25
refactor: rely on psdk storage for sidepanel preferences

ItsANameToo:
## Summary

handles https://app.clickup.com/t/86dwb56m6 by moving preference storage to the PSDK. Also moves the logic to a hook and introduces an enum for the `single` and `multiple` view options

## Checklist

-   [ ] Ready to be merged

### DIFF
--- a/src/domains/portfolio/hooks/use-address-panel.ts
+++ b/src/domains/portfolio/hooks/use-address-panel.ts
@@ -1,0 +1,108 @@
+import { useEffect, useState } from "react";
+
+import { AddressViewSelection } from "../components/AddressesSidePanel";
+import { Contracts } from "@ardenthq/sdk-profiles";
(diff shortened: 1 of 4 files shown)


src/domains/portfolio/hooks/use-address-panel.ts
```diff
@@ -0,0 +1,108 @@
...
+	return {
+		addressViewPreference,
+		singleSelectedAddress,
+		multiSelectedAddresses,
+		setAddressViewPreference,
+		setSingleSelectedAddress,
+		setMultiSelectedAddresses,
+		// Also expose the portfolio selected addresses for convenience
+		selectedAddresses,
+		setSelectedAddresses,
```
ItsANameToo, line 106: actually may not need to expose these anymore due to some refactor

PR #1121, opened 2025-04-02
refactor: move psdk helpers to library folder

ItsANameToo:
this moves the psdk `sdk-helpers` to an interal `lib` folder. Only the methods that we use in Vault have been moved over, including their internal dependencies (those are not exported at the root level at the moment)

part of https://app.clickup.com/t/86dw5864z


ItsANameToo:
want to see whether the tests also run into the `assert` issues here or if it's a local issue on my side

ItsANameToo:
local issue was due to using node 22, where the json files from the mainsail packages cause issues on the `evm.14` version (using `assert` instead of `with`) . This will be resolved with `evm.16` which we'll update to later

.github/workflows/test.yml
```diff
@@ -118,6 +118,19 @@ jobs:
                   max_attempts: 2
                   command: pnpm test:coverage src/app/contexts/Navigation/*test*
 
+            - name: Test Helpers (Lib)
+              env:
+                  COVERAGE_THRESHOLD_LINES: 100
+                  COVERAGE_THRESHOLD_FUNCTIONS: 100
+                  COVERAGE_THRESHOLD_STATEMENTS: 100
+                  COVERAGE_THRESHOLD_BRANCHES: 100
+                  COVERAGE_INCLUDE_PATH: src/app/lib/helpers/*test*
```
ItsANameToo, line 127: these don't actually do anything as coverage is reported as 0

.github/workflows/test.yml
```diff
@@ -436,36 +449,36 @@ jobs:
...
-                with:
-                    timeout_minutes: 10
-                    max_attempts: 1
-                    command: pnpm test:coverage src/domains/portfolio --exclude src/domains/portfolio/components/ImportWallet/Ledger
+            - uses: actions/checkout@v4
+              with:
+                  ref: ${{ github.head_ref }}
+            - uses: pnpm/action-setup@v4
+              with:
+                  version: 9
```
ItsANameToo, line 457: reminder to update ci to pnpm 10 after this PR is merged

eslint.config.mjs
```diff
@@ -177,9 +178,10 @@ export default [
 			"sonarjs/cognitive-complexity": "error",
 			"sonarjs/no-all-duplicated-branches": "error",
 			"sonarjs/no-collapsible-if": "error",
-			"sonarjs/no-duplicate-string": ["error", { "threshold": 5 }],
+			"sonarjs/no-duplicate-string": ["error", { threshold: 5 }],
 			"sonarjs/no-identical-expressions": "error",
 			"sonarjs/no-identical-functions": "error",
+			"sonarjs/no-one-iteration-loop": "off",
```
ItsANameToo, line 184: let me recheck if this is still needed, don't think so after adjusting the uniq/uniq-by files

package.json
```diff
@@ -221,5 +225,10 @@
 				"react-zendesk"
 			]
 		}
+	},
+	"pnpm": {
```
ItsANameToo, line 229: required for pnpm 10 to build bcrypto

samharperpittam (approved):

PR #1125, opened 2025-04-02
fix: refetch nonce when redoing same transaction

alfonsobries:
## Summary
https://app.clickup.com/t/86dwda3f4 and https://app.clickup.com/t/86dwbjhkf

Depends on https://github.com/ArdentHQ/platform-sdk/pull/208

## Checklist

-   [ ] Ready to be merged


alfonsobries:
@ItsANameToo FYI

Now that I think more about this solution, I believe the right approach might actually be not to cache the wallets at all, or maybe just reduce the TTL to a couple of seconds (to prevent multiple simultaneous requests).

I was considering the scenario where the nonce changes because a transaction was made from another instance of the wallet or something like that. In that case, there's no way to know if we should invalidate the wallet cache.

I'm leaving the PR as-is for now, but I wanted to mention this so we can discuss it tomorrow.

alfonsobries:
Also cache may be related to this ticket https://app.clickup.com/t/86dwbjhkf

ItsANameToo:
> @ItsANameToo FYI
> 
> Now that I think more about this solution, I believe the right approach might actually be not to cache the wallets at all, or maybe just reduce the TTL to a couple of seconds (to prevent multiple simultaneous requests).
> 
> I was considering the scenario where the nonce changes because a transaction was made from another instance of the wallet or something like that. In that case, there's no way to know if we should invalidate the wallet cache.
> 
> I'm leaving the PR as-is for now, but I wanted to mention this so we can discuss it tomorrow.

The way I see the cache being used is as follows:
- in case of having multiple addresses, you want to have their general data to avoid needless requests when navigating between addresses or when the sidepanel opens and shows limited address data. Having cached data used there and updated at a lower frequency avoids a lot of requests for data the user isn't necessarily interested in
- when interacting with an actual address, for example to send a transaction, we will always want to have the latest data. We can of course clear a cache when an action is carried out through Vault, but as you said that does not handle changes from the outside (and in addition if there is a delay on the network due to a validator not forging it could also mean that the cache is repopulated before the changes are applied on-chain). To counter this, I'd argue that in transaction flows we always fetch the latest address details to avoid this. This could be at the start of the flow, meaning that if a value is updated in between steps (e.g. nonce), the user can use the `back` button at the last step where it would re-trigger fetching fresh data as well

alfonsobries:
@ItsANameToo updated this PR to clear the wallet details request cache before sending a transaction, so it always fetches a fresh nonce

That makes the part for "refetching the nonce" (which is more like clearing the cache) when pressing the back button unnecessary, since it now always loads a new nonce.

Note: the api does not immediately returns a fresh nonce which means tat if the user makes two transactions in a short time it may see the nonce error which the retry button should solve

ItsANameToo:
@alfonsobries can be updated with psdk alpha.67

alfonsobries:
@ItsANameToo updated ✅

samharperpittam (approved):
Looking much better now :oleglala:

ItsANameToo (approved):

PR #1130, opened 2025-04-04
refactor: migrate `sdk-mainsail` into arkvault

goga-m:
Moves `sdk-mainsail` package internally
Closes https://app.clickup.com/t/86dwf5ghn

Part of https://app.clickup.com/t/86dw5864z

## Checklist

-   [ ] Ready to be merged


package.json
```diff
@@ -46,6 +46,7 @@
 		"@ardenthq/arkvault-url": "^1.2.0",
 		"@ardenthq/sdk": "2.0.0-alpha.66",
 		"@ardenthq/sdk-ark": "2.0.0-alpha.66",
+		"@ardenthq/sdk-cryptography": "2.0.0-alpha.66",
```
ItsANameToo, line 49: can we avoid this? we already have parts of the crypto sdk moved into vault so it will be better to use those or add the missing features to it

eslint.config.mjs
```diff
@@ -186,6 +186,7 @@ export default [
 			"sonarjs/no-use-of-empty-return-value": "error",
 			"sonarjs/no-nested-template-literals": "warn",
 			"sonarjs/prefer-single-boolean-return": "off",
+			"sonarjs/no-one-iteration-loop": "off",
```
ItsANameToo, line 189: if this is due to one of the helper files erroring on it, I resolved that in my PR

ItsANameToo (changes requested):
@goga-m can you check which of the files in the `crypto` folder are still used by mainsail? things like `schnorr` should be unused and it will likely save some polyfills when cryptography gets limited 🤔

eslint.config.mjs
ItsANameToo: reason there was not due to the rule itself, but some unnecessary typescript types that their parser didn't handle. Alternatively you can ignore it in the file itself if it remains

ItsANameToo: sounds fair @goga-m , schnorr will be gone by then and others that cause issues might as well be

PR #1156, opened 2025-04-11
fix: handle abort errors on unique name validation

patricio0312rev:
# [[username] incorrectly reporting username has been taken](https://app.clickup.com/t/86dwfwd3h)

## Summary

- Aborted requests were being handled as regular errors when validating the registration username, causing several options to be marked as "previously registered" when they weren't. For this, we're now adding a checker for aborted request and letting them pass to prevent this kind of conflicts.

## Checklist

-   [ ] Ready to be merged


samharperpittam (approved):

src/domains/transaction/validations/UsernameRegistration.ts
```diff
@@ -42,7 +42,11 @@ export const usernameRegistration = (t: any) => ({
 			unique: debounceAsync(async (value) => {
 				try {
 					await usernameExists(network, value, controller);
-				} catch {
+				} catch (error) {
+					if (error && error.name === "AbortError") {
+						return true;
+					}
+
 					return t("COMMON.VALIDATION.EXISTS", { field: t("COMMON.USERNAME") });
```
ItsANameToo, line 50: wouldn't it be better to check for the error you expect instead and then return it as "already exists" instead of ignoring one specific error while others may still cause false positives? 🤔

ItsANameToo (changes requested):

ItsANameToo (approved):

PR #1159, opened 2025-04-14
fix: handle contract deployments in transactions table

shahin-hq:
## Summary
Closes: https://app.clickup.com/t/86dvabprn

```
0x47ea9bAa16edd859C1792933556c4659A647749C
```

image

image

## Checklist

-   [ ] Ready to be merged


src/domains/transaction/hooks/use-transaction-types.ts
```diff
@@ -11,10 +11,10 @@ export const useTransactionTypes = ({ wallets = [] }: TransactionTypeProperties
 	const { t } = useTranslation();
 
 	const transactionTypes: Record<string, { label: string }> = {
-		delegateRegistration: {
+		validatorRegistration: {
```
shahin-hq, line 14: Fixing another bug: it will return `validatorRegistration` not `delegateRegistration`

ItsANameToo (changes requested):
make sure the transaction table loads the transaction properly as well

ItsANameToo (approved):

PR #1168, opened 2025-04-16
fix: increase gas limit for multipayments

patricio0312rev:
# [[multipay] transactions fail by default](https://app.clickup.com/t/86dwh10p8)

## Summary

- Multipayment transactions constantly started failing providing errors related to the gas limit being lower than expected for the pre-verification of the transaction. This amount, fluctuated betweek 22000 and 23000, reason why in this PR, I am increasing the default gas limit for multipayments to 24 000.

image

image

## Checklist

-   [ ] Ready to be merged


ItsANameToo (changes requested):
this doesn't fix the issue @patricio0312rev , it needs a much higher gas limit and it has to be based on the amount of recipients in the multipayment as well

ItsANameToo (approved):

PR #1170, opened 2025-04-16
fix: import address encryption tooltip

shahin-hq:
## Summary
Closes: https://app.clickup.com/t/86dwh49d7

[Screencast from 2025-04-16 16-11-02.webm](https://github.com/user-attachments/assets/9a9d2c3a-16f3-4c79-986c-76cebcb5ca2c)

## Checklist

-   [ ] Ready to be merged

### DIFF
--- a/src/domains/portfolio/components/ImportWallet/ImportDetailStep.tsx
+++ b/src/domains/portfolio/components/ImportWallet/ImportDetailStep.tsx
@@ -306,8 +306,8 @@ export const ImportDetailStep = ({
 
 	const [coin] = useState(() => profile.coins().get(network.coin(), network.id()));
 
-	const useEncryption = watch("useEncryption") as boolean;
-	const acceptResponsibility = watch("acceptResponsibility") as boolean;
+	const useEncryption = Boolean(watch("useEncryption"));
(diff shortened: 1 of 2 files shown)


src/domains/portfolio/components/ImportWallet/ImportDetailStep.tsx
```diff
@@ -306,8 +306,8 @@ export const ImportDetailStep = ({
 
 	const [coin] = useState(() => profile.coins().get(network.coin(), network.id()));
 
-	const useEncryption = watch("useEncryption") as boolean;
-	const acceptResponsibility = watch("acceptResponsibility") as boolean;
+	const useEncryption = Boolean(watch("useEncryption"));
+	const acceptResponsibility = Boolean(watch("acceptResponsibility"));
```
shahin-hq, line 310: This is to resolve error in dev console when toggling 

```
Warning: A component is changing an uncontrolled input to be controlled. This is likely caused by the value changing from undefined to a defined value, which should not happen. Decide between using a controlled or uncontrolled input element for the lifetime of the component. More info: https://reactjs.org/link/controlled-components
```

ItsANameToo (approved):

PR #1176, opened 2025-04-17
refactor: adjust custom servers to support 3 endpoints

shahin-hq:
## Summary
Closes: https://app.clickup.com/t/86dwh1ada

This PR refactors custom servers table and add a custom server form to support 3 endpoints.

[Screencast from 2025-04-23 15-22-46.webm](https://github.com/user-attachments/assets/2e658d2e-9339-4ef9-9b41-649e7183b880)

## Checklist

-   [ ] Ready to be merged


src/app/lib/sdk/helpers.ts
```diff
@@ -2,13 +2,27 @@ import { Array_ } from "@/app/lib/helpers";
...
+
+	for (const { host, name } of hosts) {
+		if (customHosts[name] === undefined) {
+			customHosts[name] = [];
+		}
+		customHosts[name].push({ host, name });
+	}
+
+	return customHosts;
+};
```
shahin-hq, line 24: Initially my idea was to adjust `Host` to store `publicHost`, `txHost` and `evmHost` together, but that didn't really work out, I could bend types and method signatures to match the new structure, but that would result many changes, that is why I decided to keep the original structure and add 3 hosts with different types. This function groups the custom servers so we can display all hosts in a single row, it uses `name` as an `id` to group

https://github.com/ArdentHQ/arkvault/pull/1176/files#diff-faf9fe9fd1102e964fe0b9c145f276d61609cee209ed2b6035b13036fdd88cf9R26

src/domains/setting/pages/Servers/hooks/use-hosts.ts
```diff
@@ -1,38 +1,78 @@
...
+import { NetworkHostType } from "@/app/lib/sdk/network.models";
+import { HostSet } from "@/app/lib/profiles/host.repository.contract";
 
-const findNetworkIndex = (profile: Contracts.IProfile, normalizedNetwork: NormalizedNetwork) => {
-	const { network, name, address, serverType } = normalizedNetwork;
+const findNetworkIndex = (profile: Contracts.IProfile, normalizedNetwork: NormalizedNetwork, type: NetworkHostType) => {
+	const { network, name, publicApiEndpoint, evmApiEndpoint, transactionApiEndpoint } = normalizedNetwork;
 	const networkId = network.id();
 	const parts = networkId.split(".");
-	const hosts = profile.hosts().all()[parts[0]][parts[1]] || [];
```
shahin-hq, line 8: This was throwing error when `profile.hosts().all()` returns an empty array

```
Uncaught (in promise) TypeError: Cannot read properties of undefined (reading 'devnet')
```
You can see the error here https://arkvault-git-refactor-sdk-profiles-ardent.vercel.app/

samharperpittam (changes requested):
when adding a server, can we change the description text in the modal to this to match designs:

`To add a custom peer, enter valid endpoints for the selected network.`


Current text mentions multisig which is redundant for now. 

image

samharperpittam (changes requested):
Modal is missing green ticks when an endpoint is validated

image

samharperpittam (changes requested):
Validation seems a little wonky

See video

https://github.com/user-attachments/assets/a9770d0e-badf-4156-bbf8-e31b27010b82

shahin-hq:
@samharperpittam should be good now, please check again

samharperpittam (approved):
Approved

There's a bit of weirdness when editing a peer but will turn that into a new task.

ItsANameToo (approved):

PR #1189, opened 2025-04-25
refactor: rename delegates to validators 

patricio0312rev:
# [[general] handle delegates -> validators change](https://app.clickup.com/t/86dwk30yb)

## Summary

- Regarding the `lib` packages:
  - `delegate.service` has been renamed to `validator.service`, and all the properties and methods inside have also been updated to follow this logic.
  - Types, methods, interfaces and constants referencing `delegate(s)` have also been updated and renamed to `validator(s)`
  - API call to `/delegates` is now pointing to `/validators`, as well as other child endpoints.
  - Other implementations of `delegate` have also been renamed, such as hint comments.
- Regarding React components:
  - Methods using `delegate` have been updated to support the changes in `lib`.
  - Components using `Delegate` in their name have also been updated to include `Validator`.
  - Minor changes to types to prevent ts errors.

image

## Checklist

-   [ ] Ready to be merged


samharperpittam (changes requested):
@patricio0312rev - I've tried to run this PR against my local mainsail and also in preview link.

In both scenarios, after creating a profile I get this error:

image

src/app/lib/mainsail/networks/mainsail.devnet.ts
```diff
@@ -26,15 +26,15 @@ const network: Networks.NetworkManifest = {
 			type: "full",
 		},
 		{
-			host: "https://dwallets-evm.mainsailhq.com/tx/api",
+			host: "https://dwallets-evm.mainsailhq.com/api",
```
ItsANameToo, line 29: this is not the right api

src/app/lib/mainsail/networks/mainsail.devnet.ts
```diff
@@ -26,15 +26,15 @@ const network: Networks.NetworkManifest = {
...
 			type: "tx",
 		},
 		{
 			host: "https://explorer-evm-test.mainsailhq.com",
 			type: "explorer",
 		},
 		{
-			host: "https://dwallets-evm.mainsailhq.com/evm/api",
+			host: "https://dwallets-evm.mainsailhq.com/api",
```
ItsANameToo, line 37: this is not the right api

src/app/lib/mainsail/networks/mainsail.devnet.ts
```diff
@@ -50,6 +50,7 @@ const network: Networks.NetworkManifest = {
 		wif: 186,
 	},
 	name: "Devnet",
+	//@ts-expect-error
```
ItsANameToo, line 53: why? 🤔

src/app/lib/mainsail/networks/mainsail.mainnet.ts
```diff
@@ -38,6 +38,7 @@ const network: Networks.NetworkManifest = {
 		nethash: "6e84d08bd299ed97c212c886c98a57e36545c8f5d645ca7eeae63a8bd62d8988",
 	},
 	name: "Mainnet",
+	//@ts-expect-error
```
ItsANameToo, line 41: try to avoid these lines and instead fix the error it throws

ItsANameToo (changes requested):

samharperpittam (approved):

src/app/lib/mainsail/networks/mainsail.devnet.ts
ItsANameToo: these are still wrong @patricio0312rev

src/app/lib/mainsail/networks/mainsail.devnet.ts
ItsANameToo: these are still wrong @patricio0312rev

ItsANameToo (changes requested):

ItsANameToo:
sending a vote doesn't work either on this PR @patricio0312rev , but that might be due to the wrong API being used

ItsANameToo (approved):

PR #1191, opened 2025-04-28
refactor: set single address view as default

shahin-hq:
## Summary
Closes: https://app.clickup.com/t/86dwkgk9m

This PR sets singe address view as default for the addresses side panel.

## Checklist

-   [ ] Ready to be merged


samharperpittam (changes requested):
- Create Profile
- Import Address (correctly in single mode)
- Import Second address (correctly in single mode)
- Create new address (displayed incorectly in multiview mode)

When an address is created, it should also default to single mode.

shahin-hq:
@samharperpittam I fixed it but it is not related to my changes, you can see the reported behavior in the base branch as well:

- import a wallet
- make sure it is in single view
- create a wallet - it will show `multiple addresses` label in the portfolio header

samharperpittam:
@shahin-hq 

When you create an address, does the "Viewing <address>" change to the correct address for you?

For me it seems to still display the previous address in view. 

image

And then when I open the side panel, 2 addresses have the "selected" background colour.

image

I can create a new task for this if preferred.

samharperpittam (approved):

ItsANameToo (approved):

PR #1215, opened 2025-05-08
fix: bls key never accepted

patricio0312rev:
# [[validator tx] bls key never accepted](https://app.clickup.com/t/86dwpm35y)

## Summary

- `arkvault-crypto` package has been updated to version `0.0.7`
- `publicKeyExists` method has been refactored to pass validation when a 404 error is fetched from the public keys endpoint. In case it returns this error, it means the PK is not in use, which allows its use.

image

## Checklist

-   [ ] Ready to be merged


samharperpittam (approved):

package.json
```diff
@@ -43,7 +43,7 @@
 		"sdk:check-updates": "npx npm-check-updates \"/^@ardenthq.*$/\" -u"
 	},
 	"dependencies": {
-		"@ardenthq/arkvault-crypto": "^0.0.7",
+		"@ardenthq/arkvault-crypto": "0.0.7",
```
ItsANameToo, line 46:
```suggestion
		"@ardenthq/arkvault-crypto": "^0.0.7",
```

ItsANameToo (approved):

PR #1223, opened 2025-05-13
fix: importing profile with password fails

patricio0312rev:
# [[import] importing profile with password fails](https://app.clickup.com/t/86dwq8wy3)

## Summary

- An error was being thrown in [`profile.validator.ts`](https://github.com/ArdentHQ/arkvault/blob/f343ac8d16635197af907c9cf61db1e5557e8054/src/app/lib/profiles/profile.validator.ts#L115) when the imported profile had a password and this wasn't being provided.
- This was caused due to the `validation` method from the `profile.validator` throwing errors when the file was encrypted.
- For this, the solution in this PR is proposing to check if the any "missing properties" are being reported by the validator and checking if the info in the `data` object is encrypted, so it can throw a `PasswordRequired` error.

https://github.com/user-attachments/assets/86c216ff-ed8a-4d3f-8806-fae6a696ac64

## Checklist

-   [ ] Ready to be merged


samharperpittam (approved):

src/app/lib/profiles/profile.validator.ts
```diff
@@ -115,6 +115,12 @@ export class ProfileValidator implements IProfileValidator {
 		}).validate(data, { allowUnknown: true, stripUnknown: true });
 
 		if (error !== undefined) {
+			const missingData = error.details.some((d) => d.path[0] === "data" && d.type === "any.required");
+
+			if (missingData && typeof (data as any).ciphertext === "string") {
+				throw new Error("PasswordRequired");
+			}
+
```
ItsANameToo, line 123: this feels quite obscure, any option we can check for a password prior to validating to make sure the provide password can decrypt the data?

ItsANameToo: feels a lot cleaner now

PR #1225, opened 2025-05-13
fix: mnemonic field validation

patricio0312rev:
# [[import] mnemonic field broken](https://app.clickup.com/t/86dwqjbkv)

## Summary

- A word count validation has been implemented for the Mnemonic input field.

image

## Checklist

-   [ ] Ready to be merged


ItsANameToo (changes requested):
would prefer if we could implement https://app.clickup.com/t/86dwhc1jr for a proper solution to this problem. A simple word count does not avoid people from using an invalid BIP39 phrase, which we should try to avoid

src/domains/portfolio/components/ImportWallet/ImportDetailStep.tsx
```diff
@@ -64,14 +64,14 @@ const MnemonicField = ({
...
-							findAddress,
-							network,
-							profile,
-							t,
-							value,
-						}),
+					validate: async (value: string) => {
+						const count = value.trim().split(/\s+/).length;
+						if (count !== 12 && count !== 24) {
+							return t("COMMON.INPUT_ADDRESS.VALIDATION.NOT_VALID");
```
ItsANameToo, line 70: "address is not valid" isn't true for the message either, it's the mnemonic that's invalid

ItsANameToo:
closing in favour of https://github.com/ArdentHQ/arkvault/pull/1234

PR #1255, opened 2025-05-26
fix: transfer tx sending with large amounts

shahin-hq:
## Summary
This PR fixes transfer tx sending with large amounts. Other changes/improvements:
- adjusts advanced fee up/down buttons to rely on min/max values from API
- fixes a bug related to fee mode switch - when you change gas limit in advanced view and switch to simple view, the estimated value will be set
- changes `gasPrice` & `gasLimit` to be `BigNumber` instead of `number` and refactors types to match
- adds/fixes unit tests to cover changes
- adds gas limit defaults back

Closes: 
- https://app.clickup.com/t/86dwvv8pc
- https://app.clickup.com/t/86dwvwepa

[Screencast from 2025-05-27 17-15-53.webm](https://github.com/user-attachments/assets/dd3f8c42-7c69-46e1-950d-13eddebbda85)

## Checklist

-   [ ] Ready to be merged


src/app/hooks/use-fees.ts
```diff
@@ -47,11 +47,11 @@ export function getEstimateGasParams(formData: Record<string, any>, type: string
 	const paramBuilders: Record<string, () => Omit<EstimateGasPayload, "from">> = {
 		multiPayment: () => {
 			const recipients: string[] = [];
-			const amounts: string[] = [];
+			const amounts: number[] = [];
```
shahin-hq, line 50: this should be a `number`:

image

src/app/lib/mainsail/transaction.service.ts
```diff
@@ -81,10 +81,10 @@ export class TransactionService {
 		const nonce = await this.#generateNonce(input);
 
 		const builder = TransferBuilder.new()
-			.value(UnitConverter.parseUnits(input.data.amount, "ark").valueOf())
+			.value(UnitConverter.parseUnits(input.data.amount, "ark"))
 			.to(input.data.to)
 			.nonce(nonce)
-			.gasPrice(UnitConverter.parseUnits(input.gasPrice, "gwei").toNumber())
```
shahin-hq, line 87: Passing `BigNumber` directly as the `gasPrice` supports that

src/domains/transaction/components/FeeField/FeeField.tsx
```diff
@@ -57,7 +58,8 @@ export const FeeField: React.FC<Properties> = ({ type, network, profile, ...prop
 			let gasLimit = defaultGasLimit;
 
 			try {
-				gasLimit = await estimateGas({ data: { ...getValues(), ...data }, type });
+				const gasLimitBigInt = await estimateGas({ data: { ...getValues(), ...data }, type });
+				gasLimit = BigNumber.make(gasLimitBigInt).toNumber();
```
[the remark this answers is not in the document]
shahin-hq: Makes sense, will make the change

src/app/hooks/use-fees.ts
shahin-hq: Makes sense, fixed as suggested

PR #1286, opened 2025-06-17
feat: add locked fee to validator resignation flow

alfonsobries:
## Checklist

-   [ ] Ready to be merged


alfonsobries:
> One of the tasks mentioned that the `0` scenario should be handled and is detailed in the specs. But it doesn't appear to be handled within this PR.
> 
> Essentially, when importing a genesis address such as `0xe5a97E663158dEaF3b65bBF88897b8359Dc19F81`, the "Unlocked Amount" should be 0 as those addresses registered the BLS key on Core instead of Mainsail.
> 
> The "validatorFee" for that address is 0. But the Vault UI is reporting 250ARK.

@samharperpittam this scenario si going to be handled as part of this task https://app.clickup.com/t/86dx25n1c as it requires some ongoing changes on the server

the tooltip should be working now

samharperpittam (approved):

ItsANameToo (approved):

PR #1289, opened 2025-06-18
feat: handle validator before locked fee

alfonsobries:
## Summary

https://app.clickup.com/t/86dx25n1c

## Checklist

-   [ ] Ready to be merged

### DIFF
--- a/src/app/components/PageSkeleton/PageSkeleton.tsx
+++ b/src/app/components/PageSkeleton/PageSkeleton.tsx
@@ -5,9 +5,13 @@ import { Page, Section } from "@/app/components/Layout";
 import { WelcomeSlider } from "@/domains/profile/components/WelcomeSlider/WelcomeSlider";
 import { ProfileRowSkeleton } from "@/domains/profile/components/ProfileRow/ProfileRow";
 import { Skeleton } from "@/app/components/Skeleton";
+import { useTheme } from "@/app/hooks";
+import cn from "classnames";
 
 export const PageSkeleton = () => {
 	const { t } = useTranslation();
+	const { isDarkMode } = useTheme();
+
 	return (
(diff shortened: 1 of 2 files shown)


src/domains/transaction/components/TransactionDetail/TransactionSummary/TransactionSummary.tsx
```diff
@@ -50,18 +54,16 @@ export const TransactionSummary = ({
 					</>
 				)}
 
-				{transaction.isValidatorResignation() && (
+				{transaction.isValidatorResignation() && validatorFee > 0 && (
```
ItsANameToo, line 57: can you make it show even if fee is 0? similar to how we handle it in arkscan where it shows a tooltip with more details to explain the 0 case as well as one with fees

ItsANameToo (changes requested):

src/domains/transaction/components/TransactionDetail/TransactionSummary/TransactionSummary.tsx
ItsANameToo: applies to the tx details when opening from the transaction table as well as the full flow during resignation so it's visible everywhere

alfonsobries:
@ItsANameToo fixed

samharperpittam (changes requested):
@alfonsobries 

When checking address `0x1e729Fd3Bc1D98fe32c98fE618919b79Ec878c2F` (genesis_35 which ha resigned)

The tx details modal correctly states "unlocked amount = 0". However, the tx table shows the address has gained 250ARK from the unlock when it shouldn't. 

image
image

samharperpittam (approved):

ItsANameToo (approved):

PR #1295, opened 2025-06-19
feat: highlight locked fee balance requirement

alfonsobries:
## Summary

https://app.clickup.com/t/86dx23a8n

## Checklist

-   [ ] Ready to be merged


samharperpittam (changes requested):
@alfonsobries  - found an issue with a specific scenario. Steps to reproduce:

- Import an address that has less than 250 ARK
- Import an address that has more than 250 ARK
- With the address that has less than 250 ARK, go to register a validator
- The warning will be correctly displayed
- Then within the form, change the "Sender" to the address with more than 250ARK
- The warning message for the previous address is still displayed
- If you do the above in reverse, you can bypass the warning with the address that has less than 250 ARK.

Video:

https://github.com/user-attachments/assets/cbc57111-20a0-4c62-b6fa-a50fe792dd33

ItsANameToo (approved):

PR #1302, opened 2025-06-23
fix: request loop on confirmation page

alfonsobries:
## Summary

https://app.clickup.com/t/86dx35z0q

## Checklist

-   [ ] Ready to be merged


src/app/lib/intl/datetime.ts
```diff
@@ -71,13 +69,7 @@ export class DateTime {
...
 		}
 
-		try {
-			require(`dayjs/locale/${locale}.js`);
-
-			this.#instance.locale(locale);
-		} catch {
-			console.debug(`Failed to load data for the [${locale}] locale.`);
-		}
+		this.#instance.locale(locale);
```
alfonsobries, line 72: Note: this didn't cause the loop, but the warning mentioned in the ticket. I searched through the code and we don't seem to be using any custom locale, so instead of looking for an alternative to the `require` there, I just removed it, since the other option would require a big refactor—we'd need to use async methods to do the same

PR #1312, opened 2025-06-25
feat: update core delegate with bls

alfonsobries:
## Summary

https://app.clickup.com/t/86dx23kqa and https://app.clickup.com/t/86dx58rbf

## Checklist

-   [ ] Ready to be merged


samharperpittam (changes requested):
A couple of issues with this:

1. The "Method" shows as "Contract deployment" and the tx fails when looking at mainsail scan (can view tx id `874d012897e0584843c3d0142807f975b4e4a9f0fdec7bb9aa4597eb157fcf99` on scan to see the failed tx)
image

2. The task mentions that the no funds should be locked during this process. However, if the address that is trying to update the BLS key has less than 250ARK in their address, the UI blocks them from making the tx.

samharperpittam (changes requested):
Okay so we have a little bit of inconsistency in the Update flow:

Method in TX details is "Update Validator" which is fine.
image

Method in TX Review is "Registration" but would be nice if it was "Update Validator"
image

Method in Pending/Created is "Contract Deployment". Would be good if this was Update Validator too.

image

Also a side note, it would also be good if the BLS key that appears on the TX review page was also displayed on the Pending/Created page.

alfonsobries:
@samharperpittam updated including the public key in the review page ✅

ItsANameToo:
conflicts and workflow @alfonsobries

alfonsobries:
@ItsANameToo updated ✅

samharperpittam (approved):

ItsANameToo (approved):

PR #1330, opened 2025-07-02
refactor: sidepanel back click

patricio0312rev:
# [[sidepanel] handle browser back button](https://app.clickup.com/t/86dwz4jgq)

## Summary

- "Back" click support has been added to sidepanels. This means that when you are on a side panel and using a mouse with a "back" button macro, the sidepanel will now close.
- A special implementation had to be added to the `ImportAddressSidepanel` due to its complexity, specifically because it had a `useKeydown("Enter")` handler that conflicted with browser history navigation. A flag was added just in this case to prevent this issue.

https://github.com/user-attachments/assets/3478d17c-868e-4f4a-99a4-b44d3d0d4230

## Checklist

-   [ ] Ready to be merged


samharperpittam (approved):

shahin-hq:
@patricio0312rev try the following scenarion:
- select a import option
- enter value and enable encryption
- go back with a back button
- try to enter the same value 
- you will get an error

[Screencast from 2025-07-14 14-02-19.webm](https://github.com/user-attachments/assets/15fa9885-2777-4b02-8a94-621bd411f4e9)

shahin-hq:
@patricio0312rev thank you for the work, tbh, I don't think that is a good approach we are following here. Imagine we have all the tx flows moved to sidepanel, are we going to all of these `useEffect`s and `if` checks to cover all edge cases? We need a way to bind back and continue functionality to `Back` and  `Continue` buttons, otherwise we will see bugs everywhere. Let's discuss possible solutions in Slack and see how we can handle it elegantly.

PR #1357, opened 2025-07-09
refactor: mnemonic/secret autocomplete issues

alfonsobries:
## Summary

https://app.clickup.com/t/86dx6x9pp

Note: this is an experimental solution that should be tested carefully:

I tried using `autoComplete="off"` and `"new-password"`, as well as many of the typical solutions recommended online and by AI, including adding a “honey pot” field to trick Google. I also tested some custom ideas that worked slightly better, like switching the input type from `text` to `password` when the user changes the value. That one came close, but if the user filled the field and then cleared it, autofill would show up again.

The solution I’m proposing works well but is admittedly a bit hacky, so it should be reviewed carefully. What I do is leave the actual input as a `text` input with invisible text, and place a `password` input on top that can’t be interacted with, just to simulate the masked password experience (i.e. hiding the characters).

It’s not a pretty solution, but I couldn’t find any immediate security issues with it. That said, browser behavior could change or I might be missing something. As far as I understand, `text` and `password` inputs behave the same way, except for the visual masking, so hiding the text manually should provide the same protection against prying eyes.

## Checklist

-   [ ] Ready to be merged


alfonsobries:
Ok, I've already thought of a potential security issue if you get too strict. Unlike a normal password input, you could copy the password from the input, I dont think its a big deal but also no sure

The second potential problem is that some device with a weird browser might show something unexpected in theory

So it's just a matter of weighing pros and cons.

Another alternative is the one I mentioned above, where I tried having the input as a text input and switching it to a password input once the user types something. That works, but if I delete the characters it shows autocomplete again, think it solves the problem half of the time, which is better than nothing

ItsANameToo (approved):
the only alternative I could think of for this is custom masking, but that brings its own set of cases to handle and isn't better than this approach per se. LGTM @alfonsobries

PR #1367, opened 2025-07-14
fix: adjust dashboard ledger icon

alfonsobries:
## Summary

https://app.clickup.com/t/86dx7nnkr

## Checklist

-   [ ] Ready to be merged


samharperpittam (changes requested):
image
Should be 12px of padding between the address and Ledger icon. I can't tell exactly how much i used right now but it looks like less. 

Can you double check that for me please @alfonsobries . Thanks!

PR #1369, opened 2025-07-15
fix: back button functionality on send vote error

patricio0312rev:
# [[vote] using "back" does not allow you to edit vote](https://app.clickup.com/t/86dx7a385)

## Summary

- Using "back" button when having an error on the send vote flow will redirect you to the review step.
image

## Checklist

-   [ ] Ready to be merged

### DIFF
--- a/src/domains/transaction/pages/SendVote/SendVote.tsx
+++ b/src/domains/transaction/pages/SendVote/SendVote.tsx
@@ -497,7 +497,9 @@ export const SendVote = () => {
 								<ErrorStep
 									onClose={() => navigate(`/profiles/${activeProfile.id()}/dashboard`)}
 									isBackDisabled={isSubmitting}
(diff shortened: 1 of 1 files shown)


samharperpittam (approved):

src/domains/transaction/pages/SendVote/SendVote.tsx
```diff
@@ -498,7 +508,8 @@ export const SendVote = () => {
 									onClose={() => navigate(`/profiles/${activeProfile.id()}/dashboard`)}
 									isBackDisabled={isSubmitting}
 									onBack={() => {
-										setActiveTab(Step.FormStep);
+										// Use the improved handleBack logic
+										handleBack();
```
ItsANameToo, line 512: did this change in itself not result in the expected change? with other forms we don't explicitly track the error page either, so I wonder why we need a different approach here 🤔

PR #1371, opened 2025-07-15
fix: select profile scrollabe on mobile

patricio0312rev:
# [[Select Profile] Import Profile issue on mobile](https://app.clickup.com/t/86dx5jbzw)

## Summary

- The select profile is now scrollable when the height of the screen is less than 700px, preventing issues with the fixed position.

image

## Checklist

-   [ ] Ready to be merged

### DIFF
--- a/src/domains/profile/components/Profiles/Profiles.test.tsx
+++ b/src/domains/profile/components/Profiles/Profiles.test.tsx
@@ -89,14 +89,16 @@ describe("Profiles", () => {
 
 		expect(secondSlideSkeletons?.length).toBe(2);
 	});
-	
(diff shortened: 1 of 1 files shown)


samharperpittam (approved):

ItsANameToo (changes requested):
could we handle this with css logic instead @patricio0312rev ? have the page fill and scroll automatically when there isn't enough space to show it all 🤔 I don't mind if the page itself scrolls instead of the profiles itself, since it's not a scenario a lot of users will run into but it will make everything much less complicated than the current solution. 

My idea would be to make the text that currently overlays the profiles on small screens to not be an element that overlays the other items and instead make it part of it so it naturally gets scrolled when the height is not sufficient

PR #1387, opened 2025-07-22
chore: disable ledger import option

shahin-hq:
## Checklist

-   [ ] Ready to be merged


src/domains/wallet/hooks/use-import-options.tsx
```diff
@@ -87,15 +87,17 @@ export const useImportOptions = (methods: Networks.NetworkManifestImportMethods)
 
 		let defaultOption: ImportOption | undefined;
 
-		const options: ImportOption[] = [
-			{
+		const options: ImportOption[] = [];
+
+		if (import.meta.env.VITE_LEDGER_DISABLED === "false") {
```
ItsANameToo, line 92:
```suggestion
		if (import.meta.env.VITE_LEDGER_DISABLED !== "true") {
```
that way the default will have ledger enabled (e.g. when you don't have it set in explicitly in your `.env`) while we can still disable it by setting it to true

PR #1390, opened 2025-07-22
refactor: sidepanel back click

patricio0312rev:
# [[sidepanel] handle browser back button](https://app.clickup.com/t/86dwz4jgq)

## Summary

- "Back" click support has been added to sidepanels. This means that when you are on a side panel and using a mouse with a "back" button macro, the sidepanel will now close.
- Special implementations for sidepanels that implement steps like `ImportAddressSidepanel` now use the `onBack` and `hasSteps` props to define which action the component should take depending on the current step.

https://github.com/user-attachments/assets/3478d17c-868e-4f4a-99a4-b44d3d0d4230

## Checklist

-   [ ] Ready to be merged


samharperpittam (approved):

src/app/components/SidePanel/SidePanel.tsx
```diff
@@ -32,6 +32,8 @@ interface SidePanelProps {
 	totalSteps?: number;
 	activeStep?: number;
 	footer?: React.ReactNode;
+	onBack?: () => void;
+	lastStep?: boolean;
```
ItsANameToo, line 36:
```suggestion
	isLastStep?: boolean;
```
since it's a boolean, better to use `is` or `has` in front to indicate that

ItsANameToo (changes requested):

ItsANameToo (approved):

PR #1403, opened 2025-07-28
feat: mainsail profile migration for wallets

alfonsobries:
## Summary

https://app.clickup.com/t/86dxc8u7k

## Checklist

-   [ ] Ready to be merged


src/app/lib/profiles/profile.mainsail-migrator.ts
```diff
@@ -0,0 +1,71 @@
...
+
+		return migratedWallet;
+	}
+
+	async #migrateWalletAddress(
+		profile: IProfile,
+		walletData: IProfileData["wallets"][string]["data"],
+	): Promise<IProfileData["wallets"][string]["data"]> {
+		const publicKey = walletData["PUBLIC_KEY"];
+		const wallet = await profile.walletFactory().fromPublicKey({ publicKey });
```
ItsANameToo, line 50: this does not consider the scenario where the old profile has addresses with the same public key. In that case it should only import one time and not two. Scenario is:
- import an address on Mainnet, you get an `A...` address
- import an address on Devnet with the same mnemonic, you get a `D...` address

in the old wallet these are two separate instances, but now they share the same address and should be combined

image

ItsANameToo (changes requested):

ItsANameToo (approved):

PR #1452, opened 2025-08-12
refactor: use http requests through lib services

goga-m:
## Summary
Closes https://app.clickup.com/t/86dxg708y

## Checklist

- [ ] Tests _(if necessary)_
- [ ] Ready to be merged


samharperpittam:
> @samharperpittam indeed we had an outdated url for the validator bls key check (probably in feat/mainsail too). It should be fixed now. Let me know if it still persists. Ty

@goga-m  - still doesn't appear to check for the key correctly

image

I can proceed to attempt to send the trasnsaction but it fails due to "Error encountered during contract execution: Bls Key Already Registered" (error from ARK Scan)

PR #1459, opened 2025-08-13
refactor: fetch pending transactions from endpoint

patricio0312rev:
# [[general] add pending transactions to table](https://app.clickup.com/t/86dxcw3bv)

## Summary

- Pending transactions are now fetched from `/transactions/unconfirmed` endpoint
- The API call is handled in the `pending-transactions.service`
- New Pending Transactions DTO to handle the type and other properties of these transactions.

## Missing

- Query params for filtering by `recipient` and `sender`. Already in discussion.

image

## Checklist

- [ ] Tests _(if necessary)_
- [ ] Ready to be merged

### DIFF
--- a/src/app/lib/mainsail/transaction-base.dto.test.ts
+++ b/src/app/lib/mainsail/transaction-base.dto.test.ts
@@ -48,23 +48,15 @@ describe("TransactionBaseData", () => {
 	beforeEach(() => {
 		tx = new TestTransactionData();
 		commonData = {
-			// 1 ARK in base units
 			data: "0x1234567890abcdef" as any,
-
 			from: "sender_address",
-
-			// 0.1 ARK per gas unit (assuming 1e8 base units)
 			gas: 21000,
-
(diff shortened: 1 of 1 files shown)


src/domains/transaction/hooks/use-pending-transactions.ts
```diff
@@ -2,103 +2,150 @@ import { useCallback } from "react";
...
+					blockHash: () => {},
+					confirmations: () => BigNumber.make(0),
+					convertedAmount: () => 0,
+					convertedTotal: () => 0,
+					explorerLink: () => dto.getMeta<string>("explorerLink") ?? "",
+					fee: () => dto.fee(),
+					from: () => dto.from(),
+					hash: () => dto.hash(),
+					isConfirmed: () => false,
+					isFailed: () => false,
```
ItsANameToo, line 113: how come the dto itself doesn't cover these methods? it feels off having to re-define them when majority is part of the dto already

PR #1461, opened 2025-08-14
refactor: implement select recipient design changes

shahin-hq:
## Summary
Closes: https://app.clickup.com/t/86dxf9dg6

[Screencast from 2025-08-14 20-06-27.webm](https://github.com/user-attachments/assets/c69d9614-8f5c-48b4-bb9e-93bda80215f2)

## Checklist

- [ ] Tests _(if necessary)_
- [ ] Ready to be merged


ItsANameToo (changes requested):
some UI tweaks needed:
- the divider line is not visible in dark mode (same issue in dim)
image

- the sender overview with addresses does not show the "known wallets" name, while the rest of the UI does. This means that in the sender overview you will see "Address #1" but clicking it results in a different name in the recipient field afterwards. All areas where address names are shown should use the same logic for prioritising them (known wallets, network names, local names, contacts)
- the entire row has the mouse hover state, but clicking the row doesn't do anything 🤔 only when you slick on the "select" area will it actually select it. Either change the mouse hover or make the entire row select it (preferable)

- in light mode the divider should go from blue to green when it's a selected entry. Currently it remains blue
image

- couldn't find the design for this, but the label is hard to distinguish when it's selected due to grey on green
image

- is the hover styling intentionally different when hovering over a selected entry? I couldn't find this in the designs but it looks off
image

shahin-hq:
@ItsANameToo I fixed all of the noted issues, created a task for the one we discussed: https://app.clickup.com/t/86dxh5460

PR #1479, opened 2025-08-26
refactor: fetch pending transaction

patricio0312rev:
# [[general] add pending transactions to table](https://app.clickup.com/t/86dxcw3bv)

## Summary

- Clear PR from #1459  
- Pending transactions are now fetched from `/transactions/unconfirmed` endpoint
- The API call is handled in the `pending-transactions.service`
- New Pending Transactions DTO to handle the type and other properties of these transactions.
- Implements the query params for filtering by `recipient` and `sender`. Already in discussion.

## For testing

- Currently, we are using the query params supported in `dwallets` environment, to test that, please use the following params:

```
VITE_MAINSAIL_DEVNET_FULL_HOST=https://dwallets-evm.mainsailhq.com/api
VITE_MAINSAIL_DEVNET_TX_HOST=https://dwallets-evm.mainsailhq.com/tx/api
VITE_MAINSAIL_DEVNET_EVM_HOST= https://dwallets-evm.mainsailhq.com/evm/api
```

image

## Checklist

- [ ] Tests _(if necessary)_
- [ ] Ready to be merged


src/domains/transaction/hooks/use-pending-transactions.ts
```diff
@@ -42,6 +54,43 @@ export const usePendingTransactions = (): UsePendingTransactionsReturn => {
 		[setPendingTransactions],
 	);
 
+	const addPendingTransactionFromUnconfirmed = useCallback(
```
shahin-hq, line 57: Is the `/unconfirmed` endpoint have some kind of limitation to return all unconfirmed transactions? I am asking because there are two functions exposed to fill in txs, that'd be easier to manage if we could lean on single source of truth. If the endpoint can return all unconfirmed txs  we can get rid of persisting logic - which would be very nice

src/domains/transaction/hooks/use-pending-transactions.ts
```diff
@@ -42,6 +54,43 @@ export const usePendingTransactions = (): UsePendingTransactionsReturn => {
...
+					value: input.value,
+				};
+
+				const pending: PendingTransactionData = {
+					networkId: input.networkId,
+					transaction: {
+						serialized: "",
+						signedData,
+					},
+					walletAddress: input.walletAddress,
```
ItsANameToo, line 79: here similar, the data already exists on the `input` and it feels obsolete to add it again as a meta field. If an address is necessary to determine the address it belongs to, I'd rather have it be stored by address in localstorage so we already filter on a higher level

ItsANameToo (changes requested):

samharperpittam (changes requested):
I only noticed 1 additional issue here

- Send transaction
- Before transaction confirms, close side panel
- Open transaction details for pending transaction
- The transaction will permanently remain as "Pending"

samharperpittam (approved):

ItsANameToo (changes requested):
remaining changes to tackle:
- [x] update the `from` and `to` parameters to rely on the `address` parameter instead (awaiting Mainsail update) for filtering unconfirmed transactions
- [ ] fix remaining test suite
- [x] discuss use of `ConfirmedTransactionData` for the unconfirmed endpoint. We should probably use an `UnconfirmedTransactionData` type instead to avoid confusion even though the data is similar
- [ ] discuss the use of `networkId` and `walletAddress` that is being stored together with the transaction data. Would it be more convenient to store this nested so we can easily determine the data by looking for `networkId.address.data` instead?

PR #1494, opened 2025-09-02
[wip] refactor: move validator resignation to side panel

alfonsobries:
## Summary

https://app.clickup.com/t/86dxnbx72

## Checklist

- [ ] Tests _(if necessary)_
- [ ] Ready to be merged


alfonsobries:
@goga-m Good catch!,

That error was actually present in all the transfer panels: transfer, username/validator, resignation/registration, so it would be good if everything can be quickly reviewed, should be fine now

PR #1504, opened 2025-09-09
refactor: pending transaction improvements

ItsANameToo:
## Summary

some improvements for the pending transactions PR

## Checklist

- [ ] Tests _(if necessary)_
- [ ] Ready to be merged

### DIFF
--- a/src/domains/transaction/hooks/use-profile-transactions.ts
+++ b/src/domains/transaction/hooks/use-profile-transactions.ts
@@ -510,8 +510,8 @@ export const useProfileTransactions = ({ profile, wallets, limit = 30 }: Profile
 
 				addUnconfirmedTransactionFromApi({
 					networkId: matched.networkId(),
-					walletAddress: matched.address(),
 					transaction: transaction.raw(),
+					walletAddress: matched.address(),
 				});
 			}
 		} catch (error) {
--- a/src/app/lib/mainsail/unconfirmed-transactions.service.test.ts
+++ b/src/app/lib/mainsail/unconfirmed-transactions.service.test.ts
@@ -52,11 +52,7 @@ describe("UnconfirmedTransactionsService", () => {
 	});
 
 	it("should handle undefined parameters", async () => {
-		server.use(
-			http.get(UNCONFIRMED_ENDPOINT, () => {
-				return HttpResponse.json(transactionFixture);
-			}),
-		);
+		server.use(http.get(UNCONFIRMED_ENDPOINT, () => HttpResponse.json(transactionFixture)));
 
 		const res = await service.listUnconfirmed(undefined);
 
@@ -77,7 +73,7 @@ describe("UnconfirmedTransactionsService", () => {
 			}),
 		);
 
-		const res = await service.listUnconfirmed({ limit: 50, from: [], to: [] });
+		const res = await service.listUnconfirmed({ from: [], limit: 50, to: [] });
 
 		expect(res.results.length).toEqual(4);
 	});
@@ -192,7 +188,7 @@ describe("UnconfirmedTransactionsService", () => {
 			}),
 		);
 
-		const res = await service.listUnconfirmed({ limit: 10, from: [], to: [] });
+		const res = await service.listUnconfirmed({ from: [], limit: 10, to: [] });
 
 		expect(res.results.length).toEqual(0);
 	});


package.json
```diff
@@ -46,7 +46,7 @@
 	"dependencies": {
 		"@ardenthq/arkvault-crypto": "^0.0.7",
 		"@ardenthq/arkvault-url": "^1.2.0",
-		"@arkecosystem/typescript-client": "^0.0.6",
+		"@arkecosystem/typescript-client": "^0.0.9",
```
ItsANameToo, line 49: updates typescript client so the `unconfirmed` methods use the right underlying API to avoid the previously used workaround

src/app/lib/mainsail/client.service.ts
```diff
@@ -64,6 +63,20 @@ export class ClientService {
 		);
 	}
 
+	public async unconfirmedTransactions(
```
ItsANameToo, line 66: added this to the client service so we can rely on it for fetching the unconfirmed transactions in a single place. It's using the `ConfirmedTransactionData` as it overlaps with the unconfirmed ones. We can discuss how/if we want to address this further

src/app/lib/mainsail/unconfirmed-transactions.service.ts
```diff
@@ -1,46 +1,37 @@
...
-		limit?: number;
-		offset?: number;
-		from?: string[] | string;
-		to?: string[] | string;
-	}) {
-		const normalize = (v?: string[] | string) =>
-			v == null ? undefined : (Array.isArray(v) ? v : v.split(",")).map((s) => s.trim()).filter(Boolean);
-
+	async listUnconfirmed(parameters?: { page?: number; limit?: number; from: string[]; to: string[] }) {
+		// TODO: use identifiers param instead once supported in mainsail, as it now filters based on from AND to which only works when sending to yourself
```
ItsANameToo, line 18: this requires an adjustment in the mainsail API. Once available we can use `identifiers` instead of `from` + `to` as that causes issues

src/domains/transaction/hooks/use-profile-transactions.ts
```diff
@@ -111,23 +112,20 @@ const removeConfirmedUnconfirmedTransactions = (
...
 };
 
-const getBlockTime = (wallets: Contracts.IReadWriteWallet[]): number => {
-	try {
-		const first = wallets[0];
-		if (!first) {
-			return 15_000;
-		}
+const getBlockTime = (network: Network): number => {
+	const DEFAULT_BLOCK_TIME = 8_000; // Close to expected ARK block time
```
ItsANameToo, line 116: adjusted this to a default that makes more sense

src/domains/transaction/hooks/use-profile-transactions.ts
```diff
@@ -111,23 +112,20 @@ const removeConfirmedUnconfirmedTransactions = (
...
 
-		const milestone = first.network().milestone?.();
+	try {
+		const milestone = network.milestone();
```
ItsANameToo, line 119: using the general active network now instead of relying on the first address

src/domains/transaction/hooks/use-profile-transactions.ts
```diff
@@ -111,23 +112,20 @@ const removeConfirmedUnconfirmedTransactions = (
...
 		const blockTime = get(milestone, "timeouts.blockTime") as unknown;
 		if (typeof blockTime === "number" && Number.isFinite(blockTime) && blockTime > 0) {
-			return Math.max(1000, Math.min(300_000, blockTime));
+			return Math.max(1000, blockTime); // Minimum of 1 second
```
ItsANameToo, line 122: added comment to make it more clear what is going on here

src/domains/transaction/hooks/use-unconfirmed-transactions.ts
```diff
@@ -54,28 +49,19 @@ export const useUnconfirmedTransactions = (): UseUnconfirmedTransactionsReturn =
...
-					from: input.from,
-					gasLimit: input.gasLimit,
-					gasPrice: input.gasPrice,
-					hash: input.hash,
-					nonce: input.nonce,
-					senderPublicKey: input.senderPublicKey,
-					to: input.to,
-					value: input.value,
-				};
+				const signedData: RawTransactionData = input.transaction;
```
ItsANameToo, line 54: since these are the same, no use setting them manually?

ItsANameToo:
fixing remaining test suite on the other branch

PR #1547, opened 2025-09-26
feat: minimize another panel flow

alfonsobries:
## Summary

With this change, the panel will also remain visible if the user navigates between different pages.

https://app.clickup.com/t/86dxtj0d7

## Checklist

- [ ] Tests _(if necessary)_
- [ ] Ready to be merged

### DIFF
--- a/src/domains/exchange/components/ExchangeForm/__snapshots__/ExchangeForm.test.tsx.snap
+++ b/src/domains/exchange/components/ExchangeForm/__snapshots__/ExchangeForm.test.tsx.snap
@@ -2,6 +2,545 @@
 
 exports[`ConfirmationStep > should not render without exchange transaction 1`] = `<div />`;
 
+exports[`ConfirmationStep > should render 1`] = `
+<div>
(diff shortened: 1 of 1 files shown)


samharperpittam (changes requested):
Okay this mostly looks good but there is one scenario that we may need to change.

Currently if the user does the following:

- opens a Send form
- enters detail to send form
- minimizes send form
- then navigates to Portfolio page
- Selects "Send" again
- The minimized window automatically shows.

However, after discussing it in Slack it was agreed that it would be better to treat the pressing of the Send button as a new action. So the "Pending Action in Progress" modal should be displayed even though the user is attempting to perform the same action that is minimized. 

Discussion here: https://helloardent.slack.com/archives/C03889W1KHT/p1759404439629039

This change would need to be applied to all actions within Vault.

alfonsobries:
@samharperpittam updated ✅

samharperpittam (approved):
Thanks @alfonsobries

PR #1548, opened 2025-09-30
fix: reject non eth based app in ledger import

goga-m:
## Summary
Closes https://app.clickup.com/t/86dxutbjv

## Checklist

- [ ] Tests _(if necessary)_
- [ ] Ready to be merged

### DIFF
--- a/src/app/lib/mainsail/client.service.ts
+++ b/src/app/lib/mainsail/client.service.ts
@@ -66,9 +66,9 @@ export class ClientService {
 	}
 
 	public async unconfirmedTransactions(
-		query?: Services.ClientTransactionsInput,
+		query: Services.ClientTransactionsInput = {},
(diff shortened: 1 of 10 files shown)


src/app/lib/mainsail/ledger.service.ts
```diff
@@ -114,6 +114,16 @@ export class LedgerService {
 		return [`0x`, r, s, v.toString(16)].join("");
 	}
 
+	public async isEthBasedApp() {
+		try {
+			const path = `m/44'/60'/0'/0/0`;
```
ItsANameToo, line 119: can we add a comment why this check works for our usecase? to avoid forgetting about it in the future. Each ledger app defines the paths that are allowed, and the old ark app only accepts 111 and 1 for mainnet and devnet so using 60 errors

ItsANameToo (approved):

PR #1766, opened 2026-03-02
fix: token count

shahin-hq:
## Summary
Closes: https://app.clickup.com/t/86e04k68p

This PR fixes displayed token count to rely on returned `totalCount` property instead of count of the fetched tokens.

## Checklist

- [ ] Tests _(if necessary)_
- [ ] Ready to be merged

### DIFF
--- a/src/app/lib/profiles/token.service.ts
+++ b/src/app/lib/profiles/token.service.ts
@@ -25,6 +25,7 @@ export class TokenService {
 			next: 0,
 			prev: undefined,
 			self: undefined,
+			totalCount: undefined,
 		});
 	}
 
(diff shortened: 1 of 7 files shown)


src/app/lib/profiles/token.service.ts
```diff
@@ -25,30 +24,10 @@ export class TokenService {
 			next: 0,
 			prev: undefined,
 			self: undefined,
+			totalCount: undefined,
 		});
 	}
 
-	/**
```
shahin-hq, line 31: Filtering should be on the API side, not in client.

ItsANameToo (approved):

src/app/lib/profiles/token.service.ts
ItsANameToo: will be covered with https://app.clickup.com/t/86e04p7ux

PR #1776, opened 2026-03-06
fix: wrong method in token transfers table

shahin-hq:
## Summary
Closes: https://app.clickup.com/t/86e0647xe

This PR adjusts token transfers table to have `Contract Deployment` method for contract deployments which were displayed as `Transfer` previously.

[Screencast from 2026-03-06 17-08-23.webm](https://github.com/user-attachments/assets/caea7eeb-2b72-4078-bb53-267dba60c9ff)

## Checklist

- [ ] Tests _(if necessary)_
- [ ] Ready to be merged


src/app/lib/mainsail/client.service.ts
```diff
@@ -148,6 +153,7 @@ export class ClientService {
 						status: 1,
 					},
 					...transfer,
+					to: TransactionTypeIdentifier.isTokenTransfer(transfer.functionSig) ? transfer.to : undefined,
```
shahin-hq, line 156: We need `to` to be the recipient for token transfers, otherwise `null` as it should be `Contract Deployment`

samharperpittam (approved):

goga-m (approved):

ItsANameToo (approved):

PR #1799, opened 2026-03-31
test: add coverage for `profile` domain

shahin-hq:
## Summary
Closes: https://app.clickup.com/t/86e0bj627

Adds coverage for `src/domains/profile` domain.

## Checklist

- [ ] My changes look good in both light AND dark mode
- [ ] The change is not hardcoded to a single network, but has multi-asset in mind
- [ ] I checked my changes for obvious issues, debug statements and commented code
- [ ] Documentation _(if necessary)_
- [ ] Tests _(if necessary)_
- [ ] Ready to be merged

PR #1800, opened 2026-04-01
test: add coverage for `wallets` domain

shahin-hq:
## Summary
Depends on #1799 

**Closes:** https://app.clickup.com/t/86e0bj67d

This PR adds coverage for `wallets` domain.

## Checklist

- [ ] Tests _(if necessary)_
- [ ] Ready to be merged


src/domains/wallet/pages/WalletDetails/hooks/use-wallet-options.ts
shahin-hq, line 1: Excluded musig related logic from coverage as it has hard-coded data: 

https://github.com/ArdentHQ/arkvault/blob/8fd223b8db9e5bcc3cfec0835452731007cd5109/src/domains/wallet/pages/WalletDetails/hooks/use-wallet-options.ts#L11

https://github.com/ArdentHQ/arkvault/blob/68e32dd1e1c43c0e3cb7e9965b5aec240b8c3237/src/utils/server-utils.ts#L61

ItsANameToo: will add a follow up card to remove lingering multisig logic

ItsANameToo (approved):

PR #1834, opened 2026-05-04
fix: show full value in transaction amount tooltips

goga-m:
## Summary
Closes https://app.clickup.com/t/86e11n9vv

## Checklist

- [ ] Tests _(if necessary)_
- [ ] Ready to be merged

### DIFF
--- a/src/app/lib/profiles/token.dto.ts
+++ b/src/app/lib/profiles/token.dto.ts
@@ -39,6 +39,10 @@ export class TokenDTO {
 		return this.#data.symbol.slice(0, allowedCharCount) + "…";
 	}
 
+	displayFullSymbol(): string {
+		return this.#data.symbol;
+	}
+
 	totalSupply(): BigNumber {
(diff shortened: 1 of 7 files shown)


src/app/lib/profiles/token.dto.ts
```diff
@@ -39,6 +39,10 @@ export class TokenDTO {
 		return this.#data.symbol.slice(0, allowedCharCount) + "…";
 	}
 
+	displayFullSymbol(): string {
```
shahin-hq: There is `symbol` method which has the exact implementation

samharperpittam (approved):

src/app/lib/profiles/token.dto.ts
goga-m: Indeed @shahin-hq, just pushed the fix. Thanks!

shahin-hq (approved):

PR #1841, opened 2026-05-07
fix: reduce padding between contract and copy button in token details

goga-m:
## Summary

Closes <https://app.clickup.com/t/86e12bmpa>

#### New components 

 - `WalletAddress` — a cleaner replacement for the bloated Address component. It drops the prop overload and uses only `wallet name`, `address`, and `showCopyButton` as props in one simple component.
 - `MiddleTruncation` — used under the hood by WalletAddress for more accurate and responsive middle truncation (canvas-based pixel measurement instead of the old static approach).

We only use it in `TokenDetailSidepanel` for now to test the specific bug mentioned in the task, but we can incrementally migrate other Address usages in the future.

 

## Checklist

- [ ] Tests _(if necessary)_
- [ ] Ready to be merged


ItsANameToo (changes requested):
Image
this seems to cut off part of the address' last character

goga-m:
@ItsANameToo @samharperpittam @shahin-hq I simplified the address & middle truncation component to  have css based responsive truncation. If we could have an overall look in case it doesn't break anything that would be great. Let me know. ty!

samharperpittam (changes requested):
@goga-m  - a bit confused by this one.

Task linked is to reduce padding between Contract address and copy icon. 

But it still seems to be the same as previously?

Image

goga-m:
> @goga-m - a bit confused by this one.
> 
> Task linked is to reduce padding between Contract address and copy icon.
> 
> But it still seems to be the same as previously?
> 
@samharperpittam you are right. Although I refactored the component in order to do a dynamic truncation, the approach that I took, was failing to fix the gap issue. I implemented a new more accurate component and applied it in the token details sidepanel only, so we can test again to see if it's ok.

samharperpittam (approved):

ItsANameToo (approved):

PR #1892, opened 2026-07-07
fix: persist market provider migration on existing profile

goga-m:
## Summary

Fixes coingecko  migration not persisting on old profiles for https://app.clickup.com/t/2570579/86e214efn   

## Checklist

- [ ] Tests _(if necessary)_
- [ ] Ready to be merged


src/utils/environment.ts
```diff
@@ -27,8 +27,9 @@ export const initializeEnvironment = (): Environment => {
 			"1.1.0": updateArkConstants,
 			"1.2.0": updateArkNethashes,
 			"1.3.0": updateMarketProvider,
+			"1.3.1": updateMarketProvider,
 		},
-		"1.3.0",
+		"1.3.1",
```
goga-m: @ItsANameToo I bumped the version for profiles that already run the 1.3.0 migration. Can revert it if you think it's unecessary.

samharperpittam (approved):

shahin-hq (approved):

src/utils/environment.ts
ItsANameToo: 1.3.0 would be nicer to have as only us ran it 🤔 what is necessary to remove in localstorage to rerun the migrations on demo?

goga-m: Yes we would need to clear the localstorage. Re importing the profile should work as well. I reverted the commit. Let me know if there are issues, ty.

ItsANameToo (approved):

